PR #21, opened 2021-04-06
View for Xamarin.Mac

### LINKED ISSUE #18: Xamarin.Mac interest?
Hi, great work on v2! 

I am working on a Xamarin.Mac implementation. Is there any interest of bringing xamarin.mac controls or skiaview into the project? I already have a crude working implementation.

If so, it would require a separate build target and mac tooling if the API changes (i think). The best solution would probably be to have a `LiveCharts2.Mac` package.

### LINKED ISSUE #20: Is LiveCharts2 thread-safe?
I have been playing around with lvcharts2 quite a bit now, and I love the new architecture and changes.

In my application, I am updating the ObservableCollection from a worker thread:
```csharp
int i = 0;
foreach (var item in points)
{
    ChartValues.Add(new ObservablePoint(i++, item.Value));
}
```

The most common error I see is related to collecting custom IEnumerables to arrays. Here is an example method (from `PathGeometry.cs`) which throws an out of bounds exception from the linq `ToArray(this IEnumerable obj)`:
```csharp
public override void Draw(SkiaSharpDrawingContext context)
{
    var toExecuteCommands = drawingCommands ?? (drawingCommands = commands.ToArray());
    ...
}
```

I tried fixing it with a simple lock, but it simply pushed the error to another enumeration somewhere else.

Is thread-safety within the scope of LiveCharts2, or are consumers expected to implement their own synchronization?

trympet:
Closes #18.

Implement a `CartesianChart` view for Xamarin.Mac. The solution can be built with `LiveCharts.Mac.sln`.

[Data binding](https://docs.microsoft.com/en-us/xamarin/mac/app-fundamentals/databinding) is not implemented yet. Will update once I learn how 😁


beto-rodriguez:
This is amazing, thanks!

I will review it.

To enable bindings you must register the dependency properties.

For example, the `AnimationsSpeed` property in your in your `Xamarin.Mac` implementation, you are doing:

https://github.com/trympet/LiveCharts2/blob/xamarin-mac/src/skiasharp/LiveChartsCore.SkiaSharp.Xamarin.Mac/CartesianChart.cs#L113

To make that property bindable, you must register de dependency property (or bindable properties in Xamarin):

https://github.com/beto-rodriguez/LiveCharts2/blob/master/src/skiasharp/LiveChartsCore.SkiaSharp.Xamarin.Forms/CartesianChart.xaml.cs#L129

Finally link the getter and the setter of the  the `AnimationsSpeed` property to the just registered bindable property:

https://github.com/beto-rodriguez/LiveCharts2/blob/master/src/skiasharp/LiveChartsCore.SkiaSharp.Xamarin.Forms/CartesianChart.xaml.cs#L174

And that's it, now you will be able to use bindings on your implementation

trympet:
Just got an M1 mac, so I will probably finish the view and bindings in the not-so-distant future

beto-rodriguez:
Great!

Feel free to ask me anything if you require it.

trympet:
I don't think MVVM bindings is possible in Xamarin.Mac, since Cocoa (which Xamarin mac is built on top of) is an MVC framework.

I just realized you can build Xamarin.Mac libraries on Windows if you have the Xamarin SDK installed, so you should be able to build the library. No changes required to the CI pipeline.

PR #34, opened 2021-05-04
(WPF) Fixes for use in templates

Kalcey:
Use of SetValue during initialization prevented templates from setting their own values on those properties.
Replaced property assignments with calls to SetCurrentValue, and added explicit interface implementations that use SetCurrentValue during initialization.

Added CoerceValueCallback to the Series and Axes properties to ensure they are not set to null.

### DIFF
--- a/src/LiveChartsCore/LineSeries.cs
+++ b/src/LiveChartsCore/LineSeries.cs
@@ -358,6 +358,7 @@ namespace LiveChartsCore
                     visual.Geometry.Y = y - hgs;
                     visual.Geometry.Width = gs;
                     visual.Geometry.Height = gs;
+                    visual.Geometry.RemoveOnCompleted = false;
 
                     data.TargetPoint.Context.HoverArea = new RectangleHoverArea().SetDimensions(x - hgs, y - hgs + 2 * sw, gs, gs + 2 * sw);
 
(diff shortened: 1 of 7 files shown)


Kalcey:
Closing, found an issue where geometry stopped being drawn properly under certain conditions, unsure of the exact cause as of yet.

Kalcey:
Seems the issue I encountered was unrelated to my changes, RemoveOnCompleted was not being reset on point geometry in LineSeries and RowSeries like in the other types, added that as well.

beto-rodriguez:
Thanks you so much for your work, could please help me with the template you used where you faced the issue?

Kalcey:
I added a small sample project to https://github.com/Kalcey/LiveCharts2/tree/template-sample that can illustrate both issues.
Reverting `LiveChartsCore.SkiaSharp.WPF` results in nothing being shown at all when selecting a chart.
When reverting `LiveChartsCore` the line charts' point geometry gets removed immediately when going back-and-forth between different charts.

Kalcey:
...and I found another issue thanks to this sample I haven't encountered before, switching from a line chart to a pie chart seems to delete all point geometries from the line chart.

Kalcey:
Okay found the cause of that too, the Scaler needs special handling when there is no value in a chart, otherwise things end up with NaN values and they can't recover from that.
Should I add the fix for that to this PR, or open a separate one?

beto-rodriguez:
Awesome, In this PR is OK, I will merge this one. thanks again!

Kalcey:
All right, should be all good now.

beto-rodriguez:
Looks great!

PR #61, opened 2021-06-16
StepLine Series

LabuzzMichal:
Mostly its copy-pasted from LineSeries so please take a close look :)
I feel its an overkill with creating StepLineVisualChartPoint for the purpose of StepLine Series since there is already a LineSegment which could probably be used but I got lost with abstraction layer... I would love to simplify it if you could share some tips :)

For now there are only samples for WPF available - I will add all the rest once it could be merged

### DIFF
--- a/samples/ViewModelsSamples/Index.cs
+++ b/samples/ViewModelsSamples/Index.cs
@@ -16,6 +16,13 @@
             "Lines/Custom",
             "Lines/Zoom",
 
+            "StepLines/Basic",
+            "StepLines/AutoUpdate",
+            "StepLines/Properties",
+            "StepLines/Area",
+            "StepLines/Custom",
+            "StepLines/Zoom",
+
             "StackedArea/Basic",
 
             "Bars/Basic",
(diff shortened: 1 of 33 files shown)


beto-rodriguez:
I like your idea.

So, you basically just copy pasted the `LineSeries` and used the segment you created, the `StepLineSegment`, I like it.

Let me try in your branch to simplify it, I agree with you, I think we can, I will try that the StepLineSeries inherits from LineSeries but using a different type (`IStepLineSegment`) .

LabuzzMichal:
Well, I slightly changed it as well but only some minor stuff like deleting smoothness property or deleting nullsplitting (i think its not apllicable)

beto-rodriguez:
The null splitting creates the "gaps effect", like the following image, maybe it is also useful for the step line series.

image

beto-rodriguez:
I will check it now, will come back with more news.

beto-rodriguez:
I merged it by mistake before I wanted 😄 

But everything seems right, I added multiple fixed, take a look at the git history, I also added a couple of missing features:

Now we can add gaps to step lines:

image

The library now supports stacked step lines:

image

Thanks for your help!

PR #74, opened 2021-07-05
Clean up

Orace:
List of changes by commits:
 - Optimize `using` directives.
   - Remove empty lines at start of `cs` files
   - Remove unused `using` directives
   - Sort `using` directives alphabetically
 - Made constructor methods `protected` in the abstract classes.
   - Since an abstract class can't be instantiated a `public` constructor is useless.
 - Use C# 9.0 logical patterns (`is`, `is not`) instead of equality operator for null checks.
 - Use `??=` operator when possible.


beto-rodriguez:
Thanks for that

PR #109, opened 2021-07-23
Fix MotionCanvas to support dynamic loading/unloading

### LINKED ISSUE #108: WPF MotionCanvas Doesn't Support Dynamic Loading/Unloading
**Describe the bug**
If `MotionCanvas` is dynamically unloaded and then loaded again (i.e. removed from the visual tree then added again) the associated chart does not render correctly and any further updates to the chart are ignored. 

**I have created PR https://github.com/beto-rodriguez/LiveCharts2/pull/109 to fix this issue**

**To Reproduce**
1. Check out [this branch](https://github.com/mark-monteiro/LiveCharts2/tree/unload-issue-demo) which has a modified demo example to display the issue
2. Go to the Lines/AutoUpdate sample for WPF
3. Click the "Constant changes" button to start updating the chart dynamically
4. Click the "Load/Unload Chart" twice to unload then re-load the chart
5. Notice that the chart is not rendered correctly after being reloaded and does not respond to the continues series updates in the view model

**Expected behavior**
The chart should be able to be re-loaded and render the same as before it was unloaded

**Screenshots**
N/A

**Additional context**
N/A

mark-monteiro:
Make sure `MotionCanvas` is correctly re-initialized after being unloaded then re-loaded.

Fixes #108

### DIFF
--- a/src/skiasharp/LiveChartsCore.SkiaSharp.WPF/MotionCanvas.cs
+++ b/src/skiasharp/LiveChartsCore.SkiaSharp.WPF/MotionCanvas.cs
@@ -55,7 +55,7 @@ namespace LiveChartsCore.SkiaSharpView.WPF
         /// </summary>
         public MotionCanvas()
         {
-            CanvasCore.Invalidated += OnCanvasCoreInvalidated;
+            Loaded += OnLoaded;
             Unloaded += OnUnloaded;
         }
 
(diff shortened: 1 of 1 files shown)


beto-rodriguez:
Makes sense, thanks.

This issue could also be present in the rest of the platforms, we have to check that also.

Issue #100, opened 2021-07-22
Selection with shared X axis


DevHasher:
I m trying shared X axis, and the selection doesnt seem proper
Setting on xaml
 `<lvc2:CartesianChart Name="Chart" Series="{Binding Series}" XAxes="{Binding SharedXAxis}" YAxes="{Binding YAxes}" LegendPosition="Right" LegendBackground="White" ZoomMode="X"></lvc2:CartesianChart>`

And Shared X axis declaration & definition in xaml.cs
```
  public Axis[] SharedXAxis { get; set; }

 SharedXAxis = new Axis[] 
            {
                new Axis
                {
                    Labels = labelList,
                    LabelsRotation = -90
                }
            };
```
below is the image reference.

Correct selection as axis is shared
![image](https://user-images.githubusercontent.com/23658195/126638823-863f41c2-cf64-49af-8326-48bc29cf8e03.png)

Incorrect selection even though axis is shared
![image](https://user-images.githubusercontent.com/23658195/126638901-90abd3ae-048a-4246-a3dd-eb60adcfb50f.png)

Do I need to add any other setting?

beto-rodriguez:
Try forcing the `MinStep` property in your XAxis

``` c#
XAxes = new List<Axis>
            {
                new Axis
                {
                    MinStep = TimeSpan.FromDays(1).Ticks
                }
            };
```

DevHasher:
@beto-rodriguez Min step doesn't resolve the issue.
Explaining the issue again - 
The first image in the description is the expected selection - where in the data points on both series sharing the x  axis label should be selected together.
The second image shows the incorrect selection, in which 2 different x axis label data points on series are getting selected,  and this should be prevented.

To replicate, create 2 column series with shared X axis, and try it out.

LabuzzMichal:
@beto-rodriguez please take a look once you have some time: https://github.com/beto-rodriguez/LiveCharts2/pull/111

beto-rodriguez:
This issue should be fixed with the merged commit, thanks for the report @DevHasher  and the fix @LabuzzMichal.

This fix will be included in the next version of the library.

PR #111, opened 2021-07-23
Bugfix 100

LabuzzMichal:
Solving issue https://github.com/beto-rodriguez/LiveCharts2/issues/100


beto-rodriguez:
it works thanks!

I made some format changes and also extended the fix for financial and row series (the issue might also be there).

I also made a possible performance improvement, they may be cases where your code is faster than mine, but the idea was to get over the duplicated query:

``` c#
 _chartView.Series.Where(s => s.IsColumnSeries())
/// then
_chartView.Series.Where(s => !s.IsColumnSeries()) 
```

LabuzzMichal:
I am glad it does!  Dont worry about changing my code or telling its a trash whenever it is :)

PR #114, opened 2021-07-27
[bugfix] Fix crashing when window closing on WinUI

powerpdw:
I fixed crashing when window closing on WinUI.
To reproduce the bug, the chart menus located on the left must be quickly switched before the animation ends.
Then you can see the result like the attached picture.

To fix this bug, I've modified code in 2-ways.
1. In Measure() and Draw() methods, I insert codes to check whether control to draw is disposed.
2. In MotionCanvas.xaml.cs, I insert codes to stop drawing when control was unloded.

### DIFF
--- a/src/LiveChartsCore/CartesianChart.cs
+++ b/src/LiveChartsCore/CartesianChart.cs
@@ -344,6 +344,8 @@ namespace LiveChartsCore
         /// <returns></returns>
         protected override void Measure()
         {
+            if (_chartView.ControlSize.IsEmpty) return;
+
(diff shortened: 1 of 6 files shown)


beto-rodriguez:
Thanks for your help!

This issue seems to be gone.

I just cleaned how the updates come when there are multiple threads and now this issue seems to be gone. 

Feel free to let me know if this issue persists with the current changes in the library.

powerpdw:
Thank you. We will let you know if a similar problem occurs during use.

PR #117, opened 2021-07-28
Scale `Section` coordinates correctly

mark-monteiro:
The existing cast to float was unnecessary since the `ToPixels()` method accepts a `double` value. The loss of precision from this unnecessary cast was causing incorrect scaling for some values


beto-rodriguez:
Nice catch, thanks again

PR #121, opened 2021-07-30
Fix onepoint display

powerpdw:
I've fixed diaplay bug on one-point display.

### [Before]
before

###[After]
after

### DIFF
--- a/src/LiveChartsCore/CartesianChart.cs
+++ b/src/LiveChartsCore/CartesianChart.cs
@@ -473,7 +473,6 @@ namespace LiveChartsCore
                             axis.VisibleDataBounds.Max = axis.VisibleDataBounds.Max + c;
                         }
 
-
                         var ns = axis.GetNameLabelSize(this);
                         var s = axis.GetPossibleSize(this);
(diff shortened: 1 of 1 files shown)


beto-rodriguez:
Thanks again for this PR.

I am not able to reproduce this case.

I see the current version works properly when there is only one point in a line series, am I missing something?

powerpdw:
I tested [Scatter]-[Basic] sample.
In ViewModel.cs, The code I tested is as follows.
If you test the code below you will find the problem.

```
using LiveChartsCore;
using LiveChartsCore.Defaults;
using LiveChartsCore.SkiaSharpView;
using System.Collections.Generic;
using System.Collections.ObjectModel;

namespace ViewModelsSamples.Scatter.Basic
{
    public class ViewModel
    {
        public IEnumerable<ISeries> Series { get; set; } = new ObservableCollection<ISeries>
        {
            new ScatterSeries<ObservablePoint>
            {
                Values = new ObservableCollection<ObservablePoint>
                {
                    new ObservablePoint(20, 20)                    
                }
            }
        };
    }
}
```

PR #179, opened 2021-09-06
Fixes display scaling in UWP

jnayee:
When using HIDPI monitor with display scaling set to greater than 100%, chart canvas would not scale correctly. This PR corrects this issue, see before & after pics below:-

image
image


beto-rodriguez:
@jnayee  Thanks!

thirstyape:
I believe this issue exists in the Blazor port as well. I typically have my monitor set to 150% scaling and the resulting chart looks just like the Before photo from this post.

thirstyape:
Can confirm that the Blazor port is fine as of version 2.0.0-beta.800.

PR #181, opened 2021-09-07
Fixes reattach visual scaling in UWP & WinUI

jnayee:
Fixes issue where OnPaintSurface would be called each time we toggle attach visual. This caused behaviour on HIDPI display where canvas would be scaled each time. See before and after pics below:-

image


beto-rodriguez:
Thanks again!

PR #246, opened 2021-10-15
fix Line and Scatter point mesure, TooltipFindingStrategy

wenya90:
fix chartpoint hover area for tooltip,  when mouse pointer moves near or away from.  fix TooltipFindingStrategy in FindPointsNearTo


aspion-ae:
Mets my expectation in project with LineSeries 👍 Thanks

beto-rodriguez:
Thanks!

Issue #271, opened 2021-11-04
Exception thrown in WPFSample


dmedine:
**Describe the bug**
Switching to VisualTest/ReattachVisual throws an exception here: https://github.com/beto-rodriguez/LiveCharts2/blob/master/src/LiveChartsCore/Themes/VisualsStyle.cs#L307

**To Reproduce**
1. Build and run WPFSamples in VisualStudio 2022 RC debugger
2. Click on 'VisualTest/ReattachVisual'
3. See error Exception is thrown and unhandled:

`System.InvalidCastException: 'Unable to cast object of type 'LiveChartsCore.SkiaSharpView.PolarLineSeries`1[System.Double]' to type 'LiveChartsCore.Kernel.Sketches.ILineSeries`1[LiveChartsCore.SkiaSharpView.Drawing.SkiaSharpDrawingContext]'.'`

This breaks on this line: https://github.com/beto-rodriguez/LiveCharts2/blob/master/src/LiveChartsCore/Themes/VisualsStyle.cs#L307

**Desktop (please complete the following information):**
 - Windows 10

**Additional context**
I changed the target frameworks from .NET Framework 4.6.2 (why?) to .NETCore 3.1 in LiveChartsCore, LiveChartsCore.SkiaSharpView and LiveChartsCore.SkiaSharpView.WPF

PR #272, opened 2021-11-05
fixed polar bug

dmedine:
This is a bug fix related to https://github.com/beto-rodriguez/LiveCharts2/issues/271.

The problem was actually with the implementation of PolarCharts. It seems that there was no definition of 'polar' in SeriesProperties.cs for the style object (VisualStyle.cs) to interpret _and_ that the seriesproperty in the PolarLineSeries constructor in core was being set to 'Line'. As far as I can tell, this patch corrects the issue. Polar/Basic is now working as well (same bug in effect there).


beto-rodriguez:
Thanks it makes sense

PR #297, opened 2021-11-22
Fixed calculation of tooltip position

aspion-ae:
From my point of view the current calculation of the tooltip position is not correct. I guess this calculation should give the expected tooltip position.


aspion-ae:
New findings to this issue: one of my test machines works fine with this fix, the other test machine works fine with the orginal code. **Will do more investigations...**

PR #299, opened 2021-11-22
WinUI references update

alex4998:
WinUI projects Microsoft references updated to the latest stable from experimental. SkiaSharp reference updated to 2.88.0-preview.155. Previous latest stable looks like doesn't go along with the latest Microsoft references. Compile gives Page already exists errors for all pages. Strange behavior. Not researched.


beto-rodriguez:
This could be what caused the issue, thanks

PR #421, opened 2022-03-12
Fix Persian and Arabic text(label) rendering

raminmjj:
This pull request fixes RTL and other similar languages renderings.
Before:
Before

After:
After


beto-rodriguez:
Thanks!

PR #433, opened 2022-04-05
Fix Build issue with Maui

ChrisPulman:
Rectangle has been changed to Rect in Maui


beto-rodriguez:
Thanks for the update and the PR

PR #438, opened 2022-04-12
Fixed a bug in the axis that could cause an incorrect size of area for the labels if _forceStepToMin is true

MarcF12:
The fix is similar to the code in Axis.cs line 249 and PolarAxis.cs line 223. This piece was missing in the `GetPossibleSize(Chart<TDrawingContext> chart)` method, such that the area where the labels are shown could get an incorrect size.


beto-rodriguez (approved):

beto-rodriguez:
Thanks! this was definitely missing.

PR #441, opened 2022-04-19
Uno platform

beto-rodriguez:
Adds support for Uno Platform


beto-rodriguez:
This PR will continue in a new PR, I will just add support for net 6 for now.

PR #524, opened 2022-06-14
Updated WinUI project to .NET 6.

alex4998:
It seems that all other .NET core projects were updated to .NET 6 and this one still referenced .NET 5. Fixing that. Tested resulting binaries in my WinUI app.


beto-rodriguez:
Thanks!

PR #581, opened 2022-08-14
Zoom by section

beto-rodriguez:
Zooms works exactly as it used to but now we can use the right button to select the section we are zooming to:

zoomingSection


beto-rodriguez:
This is not working yet on Mobile devices.

For mobile devices the 'ZoominSection' will be activated on double tap.

beto-rodriguez:
Actually on mobile devices the zooming is controlled by the user fingers, this seems unnecessary on mobile devices.

PR #701, opened 2022-10-20
Crosshair snapping

EniacMlezi:
Created a way for making the crosshair "snap" to the closest available ChartPoint. Can be enabled/disabled on a per axis basis

https://user-images.githubusercontent.com/15785664/196917984-4d97c5dc-eab4-40e6-be70-b40f24f90f07.mp4

### DIFF
--- a/src/LiveChartsCore/Axis.cs
+++ b/src/LiveChartsCore/Axis.cs
@@ -721,7 +721,10 @@ public abstract class Axis<TDrawingContext, TTextGeometry, TLineGeometry>
         chart.Update();
     }
 
-    private static ChartPoint? FindClosestPoint(LvcPoint pointerPosition, CartesianChart<TDrawingContext> cartesianChart, IEnumerable<ICartesianSeries<TDrawingContext>> allSeries)
+    private static ChartPoint? FindClosestPoint(
+        LvcPoint pointerPosition,
+        CartesianChart<TDrawingContext> cartesianChart,
+        IEnumerable<ICartesianSeries<TDrawingContext>> allSeries)
     {
         ChartPoint? closestPoint = null;
         foreach (var series in allSeries)
--- a/src/LiveChartsCore/Axis.cs
+++ b/src/LiveChartsCore/Axis.cs
@@ -91,6 +91,7 @@ public abstract class Axis<TDrawingContext, TTextGeometry, TLineGeometry>
     private bool _isVisible = true;
     private bool _isInverted;
     private bool _forceStepToMin;
+    private bool _crosshairSnapEnabled;
     private readonly float _tickLength = 8f;
     private readonly int _subSections = 3;
 
@@ -237,6 +238,9 @@ public abstract class Axis<TDrawingContext, TTextGeometry, TLineGeometry>
 
     /// <inheritdoc cref="ICartesianAxis{TDrawingContext}.CrosshairPadding"/>
     public Padding? CrosshairPadding { get; set; }
+
+    /// <inheritdoc cref="ICartesianAxis{TDrawingContext}.CrosshairSnapEnabled" />
+    public bool CrosshairSnapEnabled { get => _crosshairSnapEnabled; set { _crosshairSnapEnabled = value; OnPropertyChanged(); } }
 
     /// <summary>
     /// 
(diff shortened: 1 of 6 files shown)


src/LiveChartsCore/Axis.cs
```diff
@@ -611,13 +615,43 @@ public void InvalidateCrosshair(Chart<TDrawingContext> chart, LvcPoint pointerPo
...
-            x = pointerPosition.X;
+            float crosshairX;
+            if (CrosshairSnapEnabled)
+            {
+                var axisIndex = Array.IndexOf(cartesianChart.XAxes, this);
+                if (axisIndex == -1) return;
+
+                var closestPoint = FindClosestPoint(pointerPosition, cartesianChart, cartesianChart.Series.Where(s => s.ScalesXAt == axisIndex));
+
+                crosshairX = scale.ToPixels(closestPoint?.SecondaryValue ?? pointerPosition.X);
```
EniacMlezi, line 626: If no close point found, reverts to non-snapping crosshair. Do we want this implementation?

src/LiveChartsCore/Axis.cs
```diff
@@ -686,6 +720,25 @@ public void InvalidateCrosshair(Chart<TDrawingContext> chart, LvcPoint pointerPo
         chart.Update();
     }
 
+    private static ChartPoint? FindClosestPoint(LvcPoint pointerPosition, CartesianChart<TDrawingContext> cartesianChart, IEnumerable<ICartesianSeries<TDrawingContext>> allSeries)
+    {
+        ChartPoint? closestPoint = null;
+        foreach (var series in allSeries)
+        {
+            var hitpoints = series.FindHitPoints(cartesianChart, pointerPosition, TooltipFindingStrategy.CompareOnlyXTakeClosest);
```
EniacMlezi: I've tried different "CompareAllTakeClosest", but that gives weird behavior I don't completely understand

src/LiveChartsCore/Axis.cs
```diff
@@ -686,6 +720,25 @@ public void InvalidateCrosshair(Chart<TDrawingContext> chart, LvcPoint pointerPo
...
+            var hitpoint = hitpoints.SingleOrDefault();
+            if (hitpoint == null) continue;
```
EniacMlezi, line 732: No idea how "TakeClosest" can return no elements, but it happens sometimes. Therefore implement a continue here

src/LiveChartsCore/Axis.cs
```diff
@@ -611,13 +616,47 @@ public void InvalidateCrosshair(Chart<TDrawingContext> chart, LvcPoint pointerPo
         float x, y;
         if (_orientation == AxisOrientation.X)
         {
-            x = pointerPosition.X;
+            float crosshairX;
+            if (CrosshairSnapEnabled)
+            {
+                var axisIndex = Array.IndexOf(cartesianChart.XAxes, this);
+                if (axisIndex == -1) return;
```
EniacMlezi: return. or try drawing anyways?

src/LiveChartsCore/Axis.cs
```diff
@@ -611,13 +616,47 @@ public void InvalidateCrosshair(Chart<TDrawingContext> chart, LvcPoint pointerPo
...
+            x = crosshairX;
             y = yoo;
         }
         else
         {
+            float crosshairY;
+            if (CrosshairSnapEnabled)
+            {
+                var axisIndex = Array.IndexOf(cartesianChart.YAxes, this);
+                if (axisIndex == -1) return;
```
EniacMlezi: return, or try drawing anyways?

beto-rodriguez:
This looks amazing! thanks I will review it and let you know any observation.

EniacMlezi:
> This looks amazing! thanks I will review it and let you know any observation.

Yes, please do! I used FindHitPoints to get the closest ChartPoint, but I bet those calculations have already been done by the ToolTip finding logic. Maybe this functionality can be hooked into that.

src/LiveChartsCore/Axis.cs
beto-rodriguez: Maybe this is a good solution, one case would be when there are no points in the chart, I think it is a good default to follow the cursor while there is no data in the chart yet.

src/LiveChartsCore/Axis.cs
beto-rodriguez: Just let the library decide that for you, it considers more factors depending on the user settings https://github.com/beto-rodriguez/LiveCharts2/pull/701/commits/d411224a4c3424b2fb56753a1d129ae2e9942650

src/LiveChartsCore/Axis.cs
beto-rodriguez: Stange that take closes has no elements, but besides that it is better to use `FirstOrDefault` in this case:

> Single will throw an exception if it finds more than one record matching the criteria. First will always select the first record from the list

We can have ties, so for this library we need `First` instead.

src/LiveChartsCore/Axis.cs
```diff
@@ -611,13 +616,45 @@ public void InvalidateCrosshair(Chart<TDrawingContext> chart, LvcPoint pointerPo
         float x, y;
         if (_orientation == AxisOrientation.X)
         {
-            x = pointerPosition.X;
+            float crosshairX;
+            if (CrosshairSnapEnabled)
+            {
+                var axisIndex = Array.IndexOf(cartesianChart.XAxes, this);
```
beto-rodriguez, line 622: This is the only thing I don't completely like, we do not need to calculate the index, instead I think that the library should expose a read only property where we can get this value, The library internally can assign this index, there is no need to go and find it every time the crosshair is invalidated.

We can build that in a new PR, for now this seems ok.

PR #746, opened 2022-11-11
Always add Crosshair Line and Label geo to PaintTask

EniacMlezi:
Changing CrosshairPaint / Label to a different color made the Crosshair dissapear entirely. This was because _crosshairLine and _crosshairLabel were not added to the new PaintTasks. Fixed so they will always be added (the Add() is implemented using HashMap, so no duplicates)


beto-rodriguez:
Thanks for the fix!

PR #792, opened 2022-11-30
Reduce gc allocations on render loop

pkindruk:
I have been investigating performance and memory issues on my app recently and mentioned that LiveCharts allocate a lot of single use data on every update. This introduces GC pressure which can be avoided.

I created a test workload to highlight those issues. It has 14x4k LineSeries, 4k FinancialSeries, 150k FinancialSeries and 2 ScatterSeries of 8K points on both. Responsiveness in that setup is really bad. It moves, but has huge lags. Though this story is for another time.

This PR focuses on reducing GC allocation on render loop.

Let's take a look at allocations during render (GC Heap Alloc Ignore Free Stacks). I used PerfView Collect command here, app was already started and data loaded for 10-20 seconds. This highlights what types are being constantly allocated.

gc-pressure1 0

gc-pressure1 1

As we can see main allocation sources are:
 - `BezierData`
 - `RectangleHoverArea`
 - `HashSet<ChartPoint>` copy constructor

There is also a tricky case with `float is null`. For some reason JITter doesn't optimize boxing when using launching app using PerfView Run command. This doesn't reproduce on normal start (or at least I can't find evidence). However I need that fix in `MotionProperty<T>` keep recorded performance data clear. JITter should treat `static readonly bool` fields as constants at IL->Asm compilation and simplify comparison.

gc-pressure1 2

### DIFF
--- a/src/LiveChartsCore/Kernel/ChartPointCleanupContext.cs
+++ b/src/LiveChartsCore/Kernel/ChartPointCleanupContext.cs
@@ -32,7 +32,6 @@ internal struct ChartPointCleanupContext
 {
     public int ToDeleteCnt { get; private set; }
 
-
     public void RemovePoint(ChartPoint point)
     {
         if (point.RemoveOnCompleted) // protect against deleting same point again
@@ -41,7 +40,6 @@ internal struct ChartPointCleanupContext
             point.RemoveOnCompleted = false;
         }
     }
-
 
     public static ChartPointCleanupContext For(HashSet<ChartPoint> points)
     {
--- a/src/LiveChartsCore/Kernel/ChartPointCleanupContext.cs
+++ b/src/LiveChartsCore/Kernel/ChartPointCleanupContext.cs
@@ -74,7 +74,6 @@ internal class ChartPointCleanupContext
         PolarScaler scale,
         Action<ChartPoint, PolarScaler> disposeAction)
     {
-
         if (_toDeleteCount == 0) return;
 
         // It would be nice to have System.Buffers installed to use rented buffer
--- a/src/LiveChartsCore/Kernel/ChartPointCleanupContext.cs
+++ b/src/LiveChartsCore/Kernel/ChartPointCleanupContext.cs
@@ -28,38 +28,38 @@ using System.Linq;
 
 namespace LiveChartsCore.Kernel;
 
-internal struct ChartPointCleanupContext
+internal class ChartPointCleanupContext
 {
-    public int ToDeleteCnt { get; private set; }
-
-    public void RemovePoint(ChartPoint point)
-    {
-        if (point.RemoveOnCompleted) // protect against deleting same point again
-        {
-            ToDeleteCnt--;
-            point.RemoveOnCompleted = false;
-        }
-    }
+    private int _toDeleteCount;
 
     public static ChartPointCleanupContext For(HashSet<ChartPoint> points)
     {
-        InvalidateAllPoints(points);
-        return new ChartPointCleanupContext() { ToDeleteCnt = points.Count };
+        foreach (var point in points) point.RemoveOnCompleted = true;
+        return new ChartPointCleanupContext() { _toDeleteCount = points.Count };
     }
 
-    internal static void InvalidateAllPoints(HashSet<ChartPoint> points)
+    public void Clean(ChartPoint point)
     {
-        foreach (var point in points)
-            point.RemoveOnCompleted = true;
+        if (!point.RemoveOnCompleted) return;
+
+        _toDeleteCount--;
+        point.RemoveOnCompleted = false;
     }
 
-    internal static void RemoveInvalidPoints(HashSet<ChartPoint> points, IChartView chartView, Scaler primaryScale, Scaler secondaryScale, Action<ChartPoint, Scaler, Scaler> disposeAction)
+    public void CollectPoints(
+        HashSet<ChartPoint> points,
+        IChartView chartView,
+        Scaler primaryScale,
+        Scaler secondaryScale,
+        Action<ChartPoint, Scaler, Scaler> disposeAction)
     {
+        if (_toDeleteCount == 0) return;
+
         // It would be nice to have System.Buffers installed to use rented buffer
         // Or we can probably use single cached buffer since all calculations are running on GUI thread
         // At least we don't allocate when there is nothing to remove
         // And allocate only as much as we need to remove
-        var toDeletePoints = points.Where(p => p.RemoveOnCompleted).ToArray();
+        var toDeletePoints = points.Where(p => p.RemoveOnCompleted);
         foreach (var p in toDeletePoints)
         {
             if (p.Context.Chart != chartView) continue;
@@ -68,13 +68,20 @@ internal struct ChartPointCleanupContext
         }
     }
 
-    internal static void RemoveInvalidPoints(HashSet<ChartPoint> points, IChartView chartView, PolarScaler scale, Action<ChartPoint, PolarScaler> disposeAction)
+    public void CollectPoints(
+        HashSet<ChartPoint> points,
+        IChartView chartView,
+        PolarScaler scale,
+        Action<ChartPoint, PolarScaler> disposeAction)
     {
+
+        if (_toDeleteCount == 0) return;
+
         // It would be nice to have System.Buffers installed to use rented buffer
         // Or we can probably use single cached buffer since all calculations are running on GUI thread
         // At least we don't allocate when there is nothing to remove
         // And allocate only as much as we need to remove
-        var toDeletePoints = points.Where(p => p.RemoveOnCompleted).ToArray();
+        var toDeletePoints = points.Where(p => p.RemoveOnCompleted);
         foreach (var p in toDeletePoints)
         {
             if (p.Context.Chart != chartView) continue;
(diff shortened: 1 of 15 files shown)


pkindruk:
Change list:
1. `BezierData` turned to struct. Usage is limited to generator functions and stack of `LineSeries.Invalidate` and `PolarLineSeries.Invalidate` functions. Can probably go further with readonly structs.
2. `ChartPointContext.HoverArea` now initialized lazy.
3. Replaced copy logic for `Series.everFetched`. Now it uses `ChartPoint.RemoveOnCompleted` and counts number points that should be removed.

As for boxing I used approach mentioned in this [StackOverflow question](https://stackoverflow.com/questions/66702734/avoid-boxing-of-is-null).
JITter should treat static readonly bool fields as constants at IL->Asm compilation and simplify comparison.
However issues is probably still present for `Nullable<T>`. But right now there is no evidence, so I skipped it. Currently there are only 2 usages for `MotionProperty<Nullable<T>>`.

With this fix allocations during render look like this:
gc-pressure2

beto-rodriguez:
Thanks for profiling the library, your PR is for sure reducing the memory consumption, this is something that will improve for sure the library.

It is smart what you did, I was building a new HashSet containing the points in the UI, it seems that the points are allocated twice with my approach, but your approach is using the current HashSet and just removing the points that were not used, in general your alternative seems smarter.

About the bad performance you faced, I am also working in a high-performance package, it improves 
 drastically how things are render in the UI, you can get a clue [here](https://github.com/beto-rodriguez/LiveCharts2/releases/tag/v2.0.0-beta.350).

There are just a few things that I would like to change, I will commit to this PR to explain the changes.

Thanks this is definitely better.

src/LiveChartsCore/Kernel/ChartPointCleanupContext.cs
```diff
@@ -0,0 +1,85 @@
...
+            point.RemoveOnCompleted = true;
+    }
+
+    internal static void RemoveInvalidPoints(HashSet<ChartPoint> points, IChartView chartView, Scaler primaryScale, Scaler secondaryScale, Action<ChartPoint, Scaler, Scaler> disposeAction)
+    {
+        // It would be nice to have System.Buffers installed to use rented buffer
+        // Or we can probably use single cached buffer since all calculations are running on GUI thread
+        // At least we don't allocate when there is nothing to remove
+        // And allocate only as much as we need to remove
+        var toDeletePoints = points.Where(p => p.RemoveOnCompleted).ToArray();
```
beto-rodriguez: It seems that it is not necessary to materialize the enumerable, is `ToArray()` required here?

src/LiveChartsCore/Kernel/ChartPointCleanupContext.cs
```diff
@@ -0,0 +1,85 @@
...
+        }
+    }
+
+    internal static void RemoveInvalidPoints(HashSet<ChartPoint> points, IChartView chartView, PolarScaler scale, Action<ChartPoint, PolarScaler> disposeAction)
+    {
+        // It would be nice to have System.Buffers installed to use rented buffer
+        // Or we can probably use single cached buffer since all calculations are running on GUI thread
+        // At least we don't allocate when there is nothing to remove
+        // And allocate only as much as we need to remove
+        var toDeletePoints = points.Where(p => p.RemoveOnCompleted).ToArray();
```
beto-rodriguez: > It seems that it is not necessary to materialize the enumerable, is `ToArray()` required here?

Same here

src/LiveChartsCore/Kernel/ChartPointCleanupContext.cs
```diff
@@ -0,0 +1,85 @@
...
+using LiveChartsCore.Measure;
+using System.Collections.Generic;
+using System;
+using System.Linq;
+
+namespace LiveChartsCore.Kernel;
+
+internal struct ChartPointCleanupContext
+{
+    public int ToDeleteCnt { get; private set; }
```
beto-rodriguez: Maybe `ToDeleteCount` is a better name, also this should probably be private and instead of checking this property in each series, the method should verify the count, if it is zero then we could early return.

I mean, instead of calling this in each series:

```c#
if (toDeletePointsCnt != 0)
            RemoveInvalidPoints(everFetched, cartesianChart.View, primaryScale, secondaryScale, SoftDeleteOrDisposePoint);
```

We could:
```c#
private int _toDeleteCount;

internal static void RemoveInvalidPoints(HashSet<ChartPoint> points, IChartView chartView, Scaler primaryScale, Scaler secondaryScale, Action<ChartPoint, Scaler, Scaler> disposeAction)
{
    if  (_toDeleteCount == 0) return;

    var toDeletePoints = points.Where(p => p.RemoveOnCompleted);
    foreach (var p in toDeletePoints)
    {
        if (p.Context.Chart != chartView) continue;
        disposeAction(p, primaryScale, secondaryScale);
        _ = points.Remove(p);
    }
}
```

Finally on each series only we could remove the `if (toDeletePointsCnt != 0)`

src/LiveChartsCore/Kernel/ChartPointCleanupContext.cs
```diff
@@ -0,0 +1,85 @@
...
+
+    public void RemovePoint(ChartPoint point)
+    {
+        if (point.RemoveOnCompleted) // protect against deleting same point again
```
beto-rodriguez: We can early return here; it seems easier to read.

Instead of:

```c#
 public void RemovePoint(ChartPoint point)
{
    if (point.RemoveOnCompleted)
    {
        ToDeleteCnt--;
        point.RemoveOnCompleted = false;
    }
}
```

We could do:

```c#
 public void RemovePoint(ChartPoint point)
{
    if (!point.RemoveOnCompleted) return;
    
    ToDeleteCnt--;
    point.RemoveOnCompleted = false;
}
```

beto-rodriguez (changes requested):

src/LiveChartsCore/Kernel/ChartPointCleanupContext.cs
```diff
@@ -0,0 +1,85 @@
...
+
+using LiveChartsCore.Kernel.Sketches;
+using LiveChartsCore.Measure;
+using System.Collections.Generic;
+using System;
+using System.Linq;
+
+namespace LiveChartsCore.Kernel;
+
+internal struct ChartPointCleanupContext
```
beto-rodriguez: It makes more sense to make this object a `class`

beto-rodriguez:
Those comments summarize the changes I made, thanks again for this PR!

beto-rodriguez (approved):

src/LiveChartsCore/Kernel/ChartPointCleanupContext.cs
pkindruk: Generally yes [msdn article](https://learn.microsoft.com/en-us/dotnet/api/system.collections.generic.hashset-1.enumerator?view=net-7.0#remarks). You better get it back.

But for some reason code below works without exception and prints expected result on `net5.0` and `net6.0`, but throws on `net472` and `netcoreapp3.1`.
```cs
public class Data
{
    public int Id { get; set; }

    public bool ToRemove { get; set; }
}

static void Main(string[] args)
{
    var set = new HashSet<Data>();

    for (var i = 0; i < 500; i++)
    {
        var removeFlag = i % 2 == 1;
        set.Add(new Data { Id = i, ToRemove = removeFlag });
    }

    var toRemove = set.Where(d => d.ToRemove);
    foreach(var d in toRemove)
    {
        _ = set.Remove(d);
    }

    Console.WriteLine("Success");
    Console.WriteLine(set.Count);
    foreach (var d in set)
        Console.WriteLine(d.Id);
}
```

src/LiveChartsCore/Kernel/ChartPointCleanupContext.cs
pkindruk: Sounds good to me.

src/LiveChartsCore/Kernel/ChartPointCleanupContext.cs
pkindruk: Sounds good to me.

src/LiveChartsCore/Kernel/ChartPointCleanupContext.cs
pkindruk: It is a `struct`, because it is involved only in body of synchronous methods. Allocating it on heap makes no sense, though performance hit will be negligible. If you are afraid of boxing you could make it `ref struct` to get compiler error each time you try to put it on heap.

src/LiveChartsCore/Kernel/ChartPointCleanupContext.cs
pkindruk: It turned out that this is by design for newer .NET. However we target `netstandard2.0`, so we can't rely on this. There is a `RemoveWhere` method exists since .NET Framework. I propose the following:
```cs
var toDeletePoints = points.Where(p => p.RemoveOnCompleted);//.ToArray();
foreach (var p in toDeletePoints)
{
    if (p.Context.Chart != chartView) continue;
    disposeAction(p, primaryScale, secondaryScale);
    //_ = points.Remove(p);
}
_ = points.RemoveWhere(p => p.RemoveOnCompleted);
```

beto-rodriguez: Ok, I might need to revert this.

src/LiveChartsCore/Kernel/ChartPointCleanupContext.cs
beto-rodriguez: Ok, got it.

pkindruk: I just faced a very tricky bug in my code related to struct copying (someone forgot to use `ref`). I think you are right it is better to stick with `class` here. GC allocations are minimal and harder to mess up.

PR #965, opened 2023-03-22
.NET version updated to 7 in some cases as a replacement for version …

alex4998:
…6 and in other cases as an addition to existing list of frameworks depending on the previous was how it was referenced. Android minSdkVersion updated to 21 the minimum allowed in the latest version of Visual Studio at the moment. It was 21 in most of the projects but there were projects still referencing 16. Updated SDK version to 19041 to unify the number among projects. It was 19041 in some but it was lower version in others. Updated .NET Framework version to 4.7.2.


alex4998:
I'd like to mention there are interesting and useful language features introduced along with .NET 7. There are no reasons not to switch. Do you have any concerns about it?

beto-rodriguez:
Yes @alex4998 this will be merged soon, just need some time right now to review and merge this, will do soon!

Kevin-Seiden:
If I could offer a dissenting opinion, .Net 6 is still in [Long Term Support](https://dotnet.microsoft.com/en-us/platform/support/policy) and .Net 7 is able to use .Net 6 assemblies.  As this is a library package intended to be consumed by applications targeting various frameworks, there is advantage to maximizing compatibility with supported .Net versions. My recommendation would be to skip the STS .Net versions and just use the LTS, meaning update to .Net 8 in November.

beto-rodriguez:
I agree, I think we can only upgrade the samples projects, but the all the supported platforms should target the last version (6.0).

NickLammerskitten (approved):

Issue #859, opened 2022-12-29
Custom Tooltip in .700 for CartesianChart in WPF not working


andrew-baylis:
**Describe the bug**
In WPF, the Cartesian Chart does not allow the setting of custom tooltips in the constructor or XML

**To Reproduce**
Use the code and approach suggested in your online documentation for beta 700

**Expected behavior**
The new tooltip class being used. Instead, the default tooltip is always used.

**Desktop (please complete the following information):**
 - OS: Windows 11

**Additional context**
I traced this behaviour to the InitializeCore method in CartesianChart.cs within LiveChartsCore.SkiaSharpView.WPF
This method is called in the OnApplyTempate of the chart (i.e. after the constructor) and it sets the tooltip to a new SKDefaultTooltip.
`tooltip = new SKDefaultTooltip(); // Template.FindName("tooltip", this) as IChartTooltip<SkiaSharpDrawingContext>;`

I see that the legend is also created here, meaning that this cannot be overridden in the constructor either.

Moving the lines:
```
        legend = new SKDefaultLegend(); // Template.FindName("legend", this) as IChartLegend<SkiaSharpDrawingContext>;
        tooltip = new SKDefaultTooltip(); // Template.FindName("tooltip", this) as IChartTooltip<SkiaSharpDrawingContext>;
```
to the constructor allows the Tooltip to be replaced in XML or the constructor as expected.

andrew-baylis:
Alternatively, we can leave the lines in the InitializeCore method, but check to see if they have already been assigned before overwriting them:
```
legend ??= new SKDefaultLegend(); // Template.FindName("legend", this) as IChartLegend<SkiaSharpDrawingContext>;
tooltip ??= new SKDefaultTooltip(); // Template.FindName("tooltip", this) as IChartTooltip<SkiaSharpDrawingContext>;
```

beto-rodriguez:
This is already working in the latest release thanks for the report.

PR #985, opened 2023-04-20
#859 Do not override tooltip in InitializeCore method

maagy:
User wouldn't be able to set custom tooltip. Just provide default value in constructor.


beto-rodriguez:
Makes sense thanks, some previous changed broke this, I will keep this open and will merge in a future release, In general tooltips and legends need a review.

build/pack.singed.ps1
```diff
@@ -0,0 +1,7 @@
+& $PSScriptRoot\pack.ps1 Release
+
+& dotnet nuget push **/*.nupkg --api-key oy2o6gainuded3gc42sq2gddywqhhhiu3cxtns4lokf5hu --source https://api.nuget.org/v3/index.json --skip-duplicate
```
beto-rodriguez, line 3: oh.... this token was exposed... remove it.

beto-rodriguez (approved):

PR #986, opened 2023-04-22
Update Samples to CommunityToolkit.Mvvm 8.1.0. Fix Breaking Changes

Kugelhaufen:
This upgrades CommunityToolkit.Mvvm from 8.0.0-preview3 to 8.1.0.
There were some breaking changes:
[ICommand] -> [RelayCommand]
ViewModels now inherit from ObservableObject

I was kinda confused for a moment when I wanted to try the samples on lvcharts.com so I figured it would be a good idea to upgrade.


beto-rodriguez:
Thanks for the update!

beto-rodriguez (approved):

PR #1202, opened 2023-08-31
Update boxseries.md

Zottel92:
While exploring this new Chart-Type I noticed that the comment in the docs was having the wrong constructor argument order.

I really like the new Box-Plot Chart, before I had it faked with the financial Chart :laughing: 
Keep up the good work :+1:


beto-rodriguez:
Thanks! I will merge when the tests pass.

beto-rodriguez (approved):

PR #1272, opened 2023-09-22
Change MapProjector.ToMap() to avoid unnecessary memory alloc.

enjoycode:
Use LvcPoint and LvcPointD to replace float array and double array.


beto-rodriguez (approved):

beto-rodriguez:
Hello and thanks for the PR, it seems that something is not compiling :(

PR #1394, opened 2023-12-08
Fixing stuck _isInternalSet flag

BHandle:
Fixes three cases where logic can flag that internal changes are being made to ChartElement objects and escape the scope without reverting the _isInternalSet flag.  I was actually encountering this on a project where I set the stroke color of a LineSeries based on a condition in the code.  Due to the timing involved, there was a race condition where I was updating the stroke color after the CartesianChart escaped due to the empty bounds with the flag still set, preventing my change from ever being applied.


beto-rodriguez (approved):

beto-rodriguez (approved):

beto-rodriguez:
Late but merged. thanks for the PR!

> I think _isInternalSet should be wrapped in a disposable class that can be set in a using() context

That sounds amazing to prevent this kind of issues, sure a PR is welcomed! (I promise to merge sooner 🤣)

Issue #1495, opened 2024-04-26
Exception of “an excessive number of separators,” for no deep zooming


guilhermeolisi:
I’m using LiveChart2 (version 2.0.0-rc2) in an Avalonia 11 (version 11.0.9) project. My control includes 3 CartesianCharts, and I frequently need to zoom and pan these charts. However, I’m encountering a recurring exception related to “an excessive number of separators,” which causes my application to crash. This issue does not occur with deep zooming, but rather when I pan the chart or apply a moderate zoom.

The exception arises within the GetPossibleMaxLabelSize() method of the LiveChartsCore.CoreAxis class. After analyzing the source code, I attempted to circumvent the issue by setting values for MinZoomDelta and MinStep, but the problem persists. I also tried controlling the MaxLimit and MinLimit of the axis, yet the exception continues to be thrown.

I have captured the values of all axis properties and non-public members of each axis in my project at the time the exception occurred (using the debug tools of Visual Studio 2022). The data is included in the attached file of this message. Unfortunately, I am unable to pinpoint the source of the problem; the property and variable values do not seem to justify the exception. It’s possible that the instance when I retrieved the axis values differs from the moment the exception is thrown.

Any assistance would be greatly appreciated. I am very fond of LiveChart2; it’s an incredible tool for data visualization, but this exception is causing my app to crash frequently.

File with Axes data: 
[LiveChart2 Axis exception.txt](https://github.com/beto-rodriguez/LiveCharts2/files/15131607/LiveChart2.Axis.exception.txt)

              Never mind, I got the issue.

This happens because when you zoom so deep then the double type has not enough precision to handle the current range, the axis is for sure stuck at an infinite loop.

I think I can fix this internally, while this is fixed you can set a limit in the axis zoom, for example in this case I force the zoom to stop at 1ms:

```c#
        XAxes = new Axis[]
        {
            new Axis
            {
                MinZoomDelta = TimeSpan.FromMilliseconds(1).Ticks,
            }
        };
```

_Originally posted by @beto-rodriguez in https://github.com/beto-rodriguez/LiveCharts2/issues/1076#issuecomment-1629941459_

guilhermeolisi:
Good night, @beto-rodriguez,

[...]

PR #1514, opened 2024-05-17
Fix repaints causing ThrowInfiniteSeparators. 

mvphelps:
Also include axis name in exception if available.

This is in response to issue #1495. Avalonia apps can cause repaints, even on charts not visible in the UI. Since Invalidate() is bypassed in this case, _stepCount would increment and eventually throw. This PR removes _stepCount and makes the check for the number of separators within EnumerateSeparators, and still throws the same exception if exceeded.

Additionally to help users troubleshoot, if the Axis has a name, the exception now includes the name in the message. I can remove this if desired, in the event of users checking the contents of the message, this could represent an API change effectively.


mvphelps:
For resolving NETSDK1202: https://learn.microsoft.com/en-us/dotnet/core/tools/sdk-errors/netsdk1202
Seems we need to add that. Given I'm not the maintainer, would like your input on this. But totally willing to add it, not a big deal. 

Not sure what do do about NETSDK1147. Does the build just fail to install workloads sometimes?

beto-rodriguez:
Thanks for the PR, I will check this.

> Not sure what do do about NETSDK1147. Does the build just fail to install workloads sometimes?

Probably the environment where Github actions are running was updated, I also need to check that.

mvphelps:
@beto-rodriguez Is see LiveChartsCore is using the various workloads for the Target framework property. I've moved the workload installation up in the build-windows.ps1 script to prior to any compilation, since all projects had failed.

beto-rodriguez (approved):

beto-rodriguez:
This fix is definitively necessary and fixes the issue.

About the CI task, there are some projects that require a specific version of visual studio (WinUI and Uno at least), since I do not have control over which version is installed in the Github actions environment, I skipped those builds for now.

I will merge as soon the test pass.

beto-rodriguez:
Tests passed, Views are compiling... probably Maui also requires a min visual studio version, I will check that and fix it in another PR.

mvphelps:
That's great @beto-rodriguez thank you.  I will definitely contribute to your project again if I find need in the future. You are easy to collaborate with!

PR #1530, opened 2024-06-14
Update overview.md

Itshadi007:
try to create with 10  lables

### DIFF
--- a/docs/polarchart/overview.md
+++ b/docs/polarchart/overview.md
@@ -338,7 +338,7 @@ namespace ViewModelsSamples.Bars.Basic
             {
                 new PolarLineSeries&lt;double>
                 {
-                    Values = new ObservableCollection&lt;double> { 5, 2, 1, 4, 3, 3 },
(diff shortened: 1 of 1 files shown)


docs/polarchart/overview.md
```diff
@@ -371,7 +371,7 @@ namespace ViewModelsSamples.Bars.Basic
             {
                 new PolarAxis
                 {
-                    Labels = new[] { "strength", "stamina", "resistance", "power", "hit points", "mana" },
+                    Labels = new[] { "strength", "stamina", "resistance", "power", "hit points", "mana",p1,p2,p3,p10 },
```
nadav26740, line 374: what does ``p1, p2, p3, p10`` means?

nadav26740 (commented):
Would you mind to give a bit more explanation why not just stay with the current example
and also please relate to the comment I left on line 374

Itshadi007:
First of all, I apologize for this.

Your example is very accurate. I really like it. I just wanted a graph with
10 values. I did it myself, but your example is perfect as it is.

Again, I apologize to you.

Issue #1076, opened 2023-06-19
LiveChartsCore.SkiaSharpView.Avalonia 2.0.0-beta.800-11.0.0-rc1.1 freezes


Developer-Alexander:
**Describe the bug**
Avalonia app freezes, when you zoom very deeply into chart. It seems to happen only when only a few points are visible or when the visible interval on the x axis is very small.

Additional Info:

- The source collection of the chart is updated very frequently in the background.
- The source collection is used as SyncContext.
- Point type is DateTimePoint.
- ZoomMode is ZoomAndPanMode.X.
- LiveChartsCore.SkiaSharpView.Avalonia: 2.0.0-beta.800-11.0.0-rc1.1
- Avalonia: 11.0.0-rc1.1

**To Reproduce**
See attached minimal app.
[ChartTest_freeze.zip](https://github.com/beto-rodriguez/LiveCharts2/files/11785145/ChartTest_freeze.zip)

**Expected behavior**
The app should not freeze.

**Screenshots**

**Desktop (please complete the following information):**
 - OS: Windows

beto-rodriguez:
Thanks for the project.

The app is not freezing for me, it gets sluggish when you have around 20k points.

This is the expected behavior, you can remove the points that you do not longer need, you could also compute your data before adding it to reduce the number of points, or you could wait for the official release of the high-performance package, I mean in general the performance seems nice, about 20k points, we can zoom in/out but to get more performance you need to run a virtualization algorithm, for more info see: https://github.com/beto-rodriguez/LiveCharts2/releases/tag/v2.0.0-beta.350

Am I missing something?

beto-rodriguez:
Never mind, I got the issue.

This happens because when you zoom so deep then the double type has not enough precision to handle the current range.

I think I can fix this internally, while this is fixed you can set a limit in the axis zoom, for example in this case I force the zoom to stop at 1ms:

```c#
        XAxes = new Axis[]
        {
            new Axis
            {
                MinZoomDelta = TimeSpan.FromMilliseconds(1).Ticks,
            }
        };
```

beto-rodriguez:
For now there is just a warning, in the next versions of the library, we will throw when we detect this issue, A possible solution is to set the `MinZoomDelta ` to a significant value depending on the case.

[...]

PR #1535, opened 2024-06-25
Fix Issue with _stepCount Not Resetting, Causing Infinite Separator Exceptions

airtaxi:
#### Problem
In the WinUI application, there was an issue where the `_stepCount` variable was not being reset, despite the UI not visibly exceeding 10,000 steps. Over time, `_stepCount` accumulated due to continuous use, which eventually triggered `ThrowInfiniteSeparators()` method calls.

#### Solution
To address this issue, I implemented a change where `_stepCount` is reset to 0 before each logic check that involves this variable.

#### Verification
Post-implementation, the change was tested by monitoring the application for extended periods. The fix proved effective, as the [application](https://github.com/airtaxi/PinStats) no longer threw exceptions after approximately 10 minutes of uptime while showing graphs, a common occurrence prior to this update.

#### Related Issue
This fix also appears to resolve a similar issue reported in [Issue #1076](https://github.com/beto-rodriguez/LiveCharts2/issues/1076#issuecomment-1967245178), further indicating the effectiveness of this solution.

### DIFF
--- a/src/LiveChartsCore/CoreAxis.cs
+++ b/src/LiveChartsCore/CoreAxis.cs
@@ -1,4 +1,4 @@
-﻿// The MIT License(MIT)
+// The MIT License(MIT)
 //
 // Copyright(c) 2021 Alberto Rodriguez Orozco & LiveCharts Contributors
 //
@@ -307,8 +307,6 @@ public abstract class CoreAxis<TDrawingContext, TTextGeometry, TLineGeometry>
     /// <inheritdoc cref="ChartElement{TDrawingContext}.Invalidate(Chart{TDrawingContext})"/>
     public override void Invalidate(Chart<TDrawingContext> chart)
     {
-        _stepCount = 0;
-
         var cartesianChart = (CartesianChart<TDrawingContext>)chart;
 
         var controlSize = cartesianChart.ControlSize;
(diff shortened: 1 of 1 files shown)


nadav26740:
You have added `` _stepCount = 0;`` in every function that ends with             
```cs
if (_stepCount++ > 10000) ThrowInfiniteSeparators();
```
so you could technically just completely remove 
 ```cs
if (_stepCount++ > 10000) ThrowInfiniteSeparators();
```
at functions ``Invalidate``, ``GetPossibleMaxLabelSize``, ``GetPossibleSize``

**Review by nadav26740 — CHANGES_REQUESTED** [review]
You have added `` _stepCount = 0;`` in every function that ends with             
```cs
if (_stepCount++ > 10000) ThrowInfiniteSeparators();
```
so you could technically just completely remove 
 ```cs
if (_stepCount++ > 10000) ThrowInfiniteSeparators();
```
at the end of functions ``Invalidate``, ``GetPossibleMaxLabelSize``, ``GetPossibleSize``

beto-rodriguez:
Thanks for the PR @airtaxi 

I think this issue is already fixed with #1514, it will be published in the next version of the library.

airtaxi:
@beto-rodriguez Thank you for letting me know that issue has already resolved!

I'm glad to hear that.

Therefore, I'm closeing this PR.

PR #1540, opened 2024-07-12
fix: Make Uno.WinUI conditional to Uno targets

jeromelaban:
This change avoids a duplicate resources issue that has been fixed in https://github.com/microsoft/microsoft-ui-xaml/issues/8857.

While upgrading WinAppSDK would also fix the issue, it's best to remove the dependency altogether.

Without this fix, the error can appear as follows:

```
ERROR: PRI175: 0x80073b0f - Processing Resources failed with error: Duplicate Entry.
ERROR: PRI277: 0x80073b0f - Conflicting values for resource 'Files/Uno.UI.Toolkit/Themes/Generic.xbf'
```

### DIFF
--- a/src/skiasharp/LiveChartsCore.SkiaSharpView.Uno.WinUI/LiveChartsCore.SkiaSharpView.Uno.WinUI.csproj
+++ b/src/skiasharp/LiveChartsCore.SkiaSharpView.Uno.WinUI/LiveChartsCore.SkiaSharpView.Uno.WinUI.csproj
@@ -25,7 +25,10 @@
     <DocumentationFile>bin\$(Configuration)\$(TargetFramework)\$(AssemblyName).xml</DocumentationFile>
   </PropertyGroup>
   <ItemGroup>
-    <PackageReference Include="Uno.WinUI" Version="5.0.19" />
+    <PackageReference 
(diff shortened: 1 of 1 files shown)


jeromelaban:
@beto-rodriguez thanks for adjusting! So the issue here is caused by the fact that there's no .NET SDK explicitly installed, which picks up the version coming from the GA agents, and changed since the last build.

There are multiple ways to do this, but the best is to explicitly install .NET in order to only have the version that was used at the time (net6 in this case). In some cases though, even installing .NET before hand does not work, and specifying an alternate install folder can help.

beto-rodriguez (approved):

beto-rodriguez:
>So the issue here is caused by the fact that there's no .NET SDK explicitly installed

Thanks, for the tip @jeromelaban, I will improve that in another PR.

For now, I will merge this PR. thanks for the PR!

Youssef1313:
@beto-rodriguez Could we have a new release with the fix from this PR?

beto-rodriguez:
@Youssef1313 should be now available at the latest release, sorry for the delay!

https://www.nuget.org/packages/LiveChartsCore.SkiaSharpView.Uno.WinUI/2.0.0-rc2.1

PR #1626, opened 2024-10-07
perf: MAUI MotionCanvas - only fetch density from MainDisplayInfo when changed

Kay-STL:
Currently, the MAUI MotionCanvas queries MainDisplayInfo on every paint in order to fetch the screen density.
Querying MainDisplayInfo on Android (11, at least) drops into Java interop to call system services - this can take a decent chunk of time, occasionally upwards of 15% of the total paint time in the application I'm profiling.

This PR tweaks the MotionCanvas to query MainDisplayInfo once in the constructor, then register a MainDisplayInfoChanged handler to update it if truly necessary.

Tested on Android, iOS, and Windows with no issues.


beto-rodriguez:
> occasionally upwards of 15% of the total paint time in the application I'm profiling

Pff that is an amazing improvement, thanks for the PR

beto-rodriguez (approved):

PR #1662, opened 2024-10-23
Update themes

beto-rodriguez:
The library has changed a lot, there are multiple outdated objects where themes are not applied.

## What does this PR do?

This PR updates the dark and light themes, fixes issues in PolarCharts, Legends, and VisualElements.

## What it does not

LiveCharts themes must be handled manually, every supported UI framework uses different techniques to detect/apply themes, the developer must handle that according to the target framework (See installation guide in the web site).

## Some screen shots

image
image

image
image

image
image


jpageacsys:
Is there an option to only apply the theme changes to the labels and tooltips but leave the chart colors the same? I'm color blind and I have a hard time distinguishing between the pastel colors used in the dark theme. The original chart colors stood out fine in the dark theme and it was easier to distinguish between them.

JeanPhilippeLux:
Hi, is it  a way to update theme in code behind? I have an application where user can change the theme of the application.

thank you.

PR #1704, opened 2024-11-18
Fix touch gestures not working on MAUI iOS once packaged

### LINKED ISSUE #1684: Touch gestures not working on iOS
**Describe the bug**
Create an MAUI app, install the package 2.0.0-rc4.5, include the `CartesianChart` basic example in the main page.
Start the app on iOS (simulator or device).
See that touch gestures are not working (i.e. tooltip not showing / zoom not working).

**Expected behavior**
Tooltip should show

**Smartphone (please complete the following information):**
 - Device: Simulator or iPhone 13
 - OS: iOS
 - Version: 17.5

**Additional context**

To solve the problem I've added this at the top of `MauiProgram` builder method:

```cs
_ = typeof(LiveChartsCore.Behaviours.ChartBehaviour).GetProperties();
```

Why? Because from the debugger I saw that entering with F11 into `contentView.AddGestureRecognizer(GetMacCatalystLongPress(contentView));` was only hitting `AddGestureRecognizer` and not `GetMacCatalystLongPress` which oddly was the only unrecognized symbol by the debugger.

So I thought...

> Maybe the symbol was somehow messed up by the linker

And that's why I've tried (succesfully) this trick.

Please note that if I check out `LiveCharts2` repo and try out the sample app, everything works.
Also, if I reference the source project from my app (instead of the NuGet package) everything works.

So maybe it's a packaging issue.. I don't really now.

One other interesting thing, from Rider pressing `GoToDefinition` on `ChartBehaviour` only suggests .NET Standard and Android version while iOS does not show up.

I hope this helps :)

Thanks for this awesome library!

albyrock87:
Fixes #1684

- Added `Directory.Build.props` to propagate properties to all project
  - I've added only `MauiVersion` because it was one of the issue, but I suggest to bring here also `Author` and other common properties (`LangVersion`?)..
- Added `global.json` file in order to use the proper .NET + workloads when debugging and packing
- Refactored MacCatalyst/iOS `ChartBehaviour` to avoid taking a (now) useless reference to the virtual view
  - Now keeps hard references to callbacks
- Moved event registration logic to MAUI handlers to avoid memory leaks and provide proper platform-level disconnection logic
  - Now requires `.UseLiveCharts()` in MAUI program
- Fixes display info not reacting to orientation / density changes on Android
- Removed `<DebugType>embedded</DebugType>` as it makes impossible to place breakpoints while debugging

I don't have a windows machine so I cannot try/verify I haven't broke anything there.

Please merge this as soon as possible because we're stuck due to #1684.
If you have questions I'm also available on Discord (same username).

Thanks


beto-rodriguez:
Thanks for the PR!

I will review and merge this asap!

albyrock87:
Hey @beto-rodriguez, I see you rebased the PR and a few conflicts appeared.

Are you gonna fix the conflicts with dev brach or should I do it?

beto-rodriguez:
@albyrock87 I changed the base to dev, I'll try to fix conflicts with the dev branch

beto-rodriguez:
@albyrock87 I've been working on #1705, that should be almost ready, it is an important change in the library that will improve drastically the flexibility of the library to create custom drawn elements.

I should be able to review this in the next days.

global.json
```diff
@@ -0,0 +1,6 @@
+{
+  "sdk": {
+    "version": "8.0.401",
```
beto-rodriguez, line 3: Is there a particular reason to use this version?

Maybe upgrading the version to 9.x is a good idea, just want to confirm, if there is a particular reason, I can make this commit.

src/skiasharp/LiveChartsCore.SkiaSharpView.Maui/LiveChartsCoreMauiAppBuilderExtensions.cs
```diff
@@ -0,0 +1,39 @@
...
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+// SOFTWARE.
+
+using LiveChartsCore.SkiaSharpView.Maui;
+
+namespace Microsoft.Maui.Hosting;
```
beto-rodriguez, line 25: Also here, can I change the namespace to `LiveChartsCore.SkiaSharpView.Maui`?

is there a particular reason for using that ns?

beto-rodriguez:
Thanks again again for the PR! I tested on all platforms, and it works!

global.json
albyrock87: I recently asked Microsoft what we should use as library authors regarding broad compatibility.

I think that for now you can stick with 8.0.100 with roll forward latest feature, but make sure you target `net8.0-ios17.0` in the TargetFramework and not just `net8.0-ios` because that implicitly targets only the latest native version released with the latest workload (right now is 18.2).

This ensures compatibility with MAUI8 and MAUI9.

Take into consideration that many developers (like me) are still on .NET8 considering that MAUI9 is still a bit "unstable", or simply because they didn't have enough time to upgrade.

src/skiasharp/LiveChartsCore.SkiaSharpView.Maui/LiveChartsCoreMauiAppBuilderExtensions.cs
albyrock87: The reason I've used this namespace is for dev-experience.

In MauiProgram.cs you already have this namespace imported, so the intellisense will automatically suggest the extension method.

This is a common pattern used by Microsoft.Extensions.* packages.

If you don't like it, do what you want: it's not a big deal :)

beto-rodriguez: Thanks @albyrock87, I will consider both comments!

albyrock87:
@beto-rodriguez the only thing I'm sad about, which was my only ask, is that I needed this to be compatible with net8 but the library now requires net9.

I'll have to fork this and go back to net8 for broader compatibility.

Anyway, thanks for releasing the new version!

beto-rodriguez:
> net8 but the library now requires net9

That should be a mistake on my side, I can see that, [net 8 should be supported](https://www.nuget.org/packages/LiveChartsCore.SkiaSharpView.Maui/2.0.0-rc5.1#supportedframeworks-body-tab), what is the error you are getting?

beto-rodriguez:
Actually, the sample in this repo uses net 8, I am confused now 🤯

albyrock87:
@beto-rodriguez I'm sorry, it has been a busy day and I haven't actually tried out the library, I just checked the global.json and got tricked by that.

My mistake, I apologize!

albyrock87:
I've installed the new version and I can confirm everything works well, thank you for your effort of keeping compatibility with .net8, I appreciate a lot!

beto-rodriguez:
Amazing! glad to read that!

Thanks for confirming that it works!

PR #1757, opened 2025-01-07
update: remove taiwan geojson, update china geojson

GaN8373:
image


beto-rodriguez:
Hello

In a future version we will provide a way to upload the geojson file, this will allow users to fix these issues.

At this point this the maps are pulled from public databases, were me or any contributor have no power or opinion.

Thanks for the PR, but i think instead we should allow the load of geojson files.

PR #1776, opened 2025-01-19
Fixes #1419 - Adjusted X axis name label coordinate calculation when InLineNamePlacement is used

### LINKED ISSUE #1419: Multiple X-Axes using InLinePlacement and start position names on top of each other
If you create a chart and add multiple x-axes and set each axis to use InLinePlacement to true and leave AxisPosition to start, the names of each axis will be on top of each other. The issue seems to be in the Measure function inside the LiveChartsCore.CartesianCharts during the XAxes foreach loop. 

If you replace the following 

image

With 

image

Image before code change:

image

Image after code change:

image

Hopefully this can be updated. Thanks.

jlimauro:
Adjusted X axis name label coordinate calculation when InLineNamePlacement option is used. This fixes a bug in which places axis names and labels a top each other when InLineNamePlacement option was used. This fixes for both Start and End axis positions. 

These changes resolve issue: #1419

### DIFF
--- a/src/LiveChartsCore/CartesianChartEngine.cs
+++ b/src/LiveChartsCore/CartesianChartEngine.cs
@@ -576,12 +576,14 @@ public class CartesianChartEngine(
 
                     // X Bottom
                     axis.NameDesiredSize = new LvcRectangle(
-                        new LvcPoint(0, ControlSize.Height - h), new LvcSize(ns.Width, h));
+                        new LvcPoint(0, ControlSize.Height - bs - h), new LvcSize(ns.Width, h));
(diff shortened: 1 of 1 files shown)


beto-rodriguez:
Hi and thanks for the fix!

I tried this long time ago, but the problem when I tested the proposed solution on #1419 it braked other cases (don't exactly remember the case), I know Unit tests passes, this means that we need stronger unit cases. I need more time to merge this fix to ensure nothing else brakes.

beto-rodriguez:
Thanks for the PR.

I will close this, instead #2190 was merged, it includes your commit also.

Issue #1809, opened 2025-02-13
Separators alignment setting does not work for Y-Axis


jnayee:
**Describe the bug**
Setting 'SeparatorsAtCenter = false' does not work for Y-Axis.  

**To Reproduce**
Update the Bars.RowsWithLabels sample with Y-Axis as follows: 
 
    <lvc:CartesianChart Series="{Binding Series}" YAxes="{Binding YAxes}"></lvc:CartesianChart>

    public Axis[] YAxes { get; set; } =
        [
            new Axis
                {
                    Labels = ["Category 1", "Category 2", "Category 3"],
                    LabelsRotation = 0,
                    SeparatorsPaint = new SolidColorPaint(new SKColor(200, 200, 200)),
                    SeparatorsAtCenter = false,
                    ForceStepToMin = true,
                    MinStep = 1
                }
        ];

**Expected behavior**
When SeparatorsAtCenter = false, the separators should be displayed on the side, not the center.  

**Screenshots**
The current behaviour is shown below:

![Image](https://github.com/user-attachments/assets/17238915-e470-43de-a38d-6f2f90c47a94)

**Desktop (please complete the following information):**
 - OS: Win10 22H2
 - Version [2.0.0-rc5.1]

beto-rodriguez:
This is included in the latest version on NuGet, thanks for the PR!

PR #1810, opened 2025-02-13
Fixes: #1809 - Y-Axis separators alignment bug

jnayee:
Fixes: #1809

Setting 'SeparatorsAtCenter'  true / false now works as expected, see pic below:

image


beto-rodriguez:
Thanks!

beto-rodriguez (approved):

PR #1935, opened 2025-08-08
Memory leak repair

1072643169:
### Purpose
Fix memory leaks in specific scenarios

### Changes
- Added conditional checks in:
  - Chart.UpdateThrottlerUnlocked()
  - MotionCanvas.Render()

### Testing
When the CartesianChart is loaded and continuously adds DateTimePoints:
1. When CartesianChart is visible, memory fluctuates normally
2. When CartesianChart is invisible, elements such as CircleGeometry and FloatMotionProperty continue to accumulate, leading to a continuous increase in memory usage
In industrial scenarios where long-term operation and continuous data refreshing are required, the impact on memory may have unforeseeable consequences

### Evidence
1. In practical applications, the program memory grows abnormally large, and a large number of Livechartscore-related objects are found in the dump
2. In demo:
   - I added counters to CircleGeometry and FloatMotionProperty types for easy observation of issues.
   - Created multiple CartesianCharts with high-frequency DateTimePoint additions to amplify the issue
   - Demo provides two leakage scenarios. I only tested CartesianChart with DateTimePoint and TimeSpanPoint, and other types may also have such problems.

### Notes
1. This might not be the optimal solution
2. This is my first attempt to submit PR to an open source project, hoping to be helpful to the LiveCharts2 project.
3. It took a lot of effort to locate this problem, look forward to getting a response.

### DIFF
--- a/src/LiveChartsCore/Chart.cs
+++ b/src/LiveChartsCore/Chart.cs
@@ -494,6 +494,16 @@ public abstract class Chart
     protected bool SizeChanged() => _previousSize.Width != ControlSize.Width || _previousSize.Height != ControlSize.Height;
 
     /// <summary>
+    /// Is it being rendered
+    /// </summary>
+    public bool _isRendering = false;
+
+    /// <summary>
+    /// Indicates if it is the first render
+    /// </summary>
+    public bool _isFirstRender = true;
+
+    /// <summary>
     /// Called when the updated the throttler is unlocked.
     /// </summary>
     /// <returns></returns>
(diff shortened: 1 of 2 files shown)


1072643169:
[demo.zip](https://github.com/user-attachments/files/21676425/demo.zip)

beto-rodriguez:
Thanks for the time you spent investigating this issue @1072643169, based on your example, I was able to track and identify the problem, see #1944, that explains the issue better.

About this PR, I think we need a new one that implements the alternative proposed on #1944.

beto-rodriguez:
Thanks again for the sample and the research @1072643169, I will close this PR in favor of ~#1945~ #1946.

PR #1945, opened 2025-08-15
Fix 1944

### LINKED ISSUE #1944: [Avalonia] Memory leak when chart is not in the UI and the data is changing.
Based on #1935.

The sample provided #1935 by @1072643169 illustrates the memory leak, it occurs when the data is changing but the chart it is not in the UI (for example in another tab or not inside the viewport).

In the next gif (2.3MB), the geometries in the canvas are logged to the console, when the chart is scrolled outside the viewport, we can see that the geometries count increases, the good news is that as soon as the chart is back in the viewport Live Charts is able to clear the geometries and the extra allocated memory in the app.

Image

This happens because LiveCharts disposes the drawn shape when it finishes the animation, since Avalonia (in this case, but might happen on other UI frameworks as well) is not requesting a frame from LiveCharts (because the control is not in the viewport) then LiveCharts is not able to dispose this these objects.

A good solution is to detect whether the chart is in the viewport and only measure the control when it is inside the viewport, even when the data binded to the chart is changing, this will prevent the library from creating new shapes while it is not visible, and of course it more efficient since it makes no sense to measure the control when it is not in the UI.

beto-rodriguez:
fixes #1944

### DIFF
--- a/src/LiveChartsCore/CartesianChartEngine.cs
+++ b/src/LiveChartsCore/CartesianChartEngine.cs
@@ -357,9 +357,8 @@ public class CartesianChartEngine(
         }
 #endif
 
-        if (!IsLoaded) return; // <- prevents a visual glitch where the visual call the measure method
-                               // while they are not visible, the problem is when the control is visible again
-                               // the animations are not as expected because previously it ran in an invalid case.
+        if (!IsLoaded || !View.IsInViewport)
+            return;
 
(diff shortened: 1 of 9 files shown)


beto-rodriguez:
replaced by https://github.com/beto-rodriguez/LiveCharts2/pull/1946.

PR #2096, opened 2026-02-12
More tests

beto-rodriguez:


beto-rodriguez:
@copilot can you implement `actions/cache@v5.0.3` into the .github/actions/detup-dotnet-and-workloads so it caches the used workloads?

Issue #2117, opened 2026-03-17
AvaloniaUI 12.0.0 support


fntc:
**Describe the bug**
Avalonia will release their 12.0.0 Release candidate this week. I am creating this bug to track API incompatibilities.

**To Reproduce**
Reference Avalonia 12.0.0-preview2 (or RC once it is available)

**Expected behavior**
App with LiveCharts2 runs as expected.

https://docs.avaloniaui.net/docs/avalonia12-breaking-changes

fntc:
Stacktrace on Startup (Windows):

Unhandled exception. System.MissingFieldException: Field not found: 'Avalonia.Input.Gestures.PinchEvent'.
   at LiveChartsGeneratedCode.SourceGenCartesianChart..ctor()
   at LiveChartsGeneratedCode.SourceGenCartesianChart..ctor()
   at LiveChartsCore.SkiaSharpView.Avalonia.CartesianChart..ctor()

Stacktrace on Startup (Android):

03-17 11:39:14.390 E/AndroidRuntime(17740): android.runtime.JavaProxyThrowable: [System.MissingFieldException]: Field not found: 'Avalonia.Input.Gestures.PinchEvent'.
03-17 11:39:14.390 E/AndroidRuntime(17740): 	at LiveChartsGeneratedCode.SourceGenCartesianChart..ctor(Unknown Source)
03-17 11:39:14.390 E/AndroidRuntime(17740): 	at LiveChartsCore.SkiaSharpView.Avalonia.CartesianChart..ctor + 0x0(Unknown Source)

beto-rodriguez:
WIP https://github.com/Live-Charts/LiveCharts2/pull/2145

JaggerJo:
We are also waiting for v12 support :)

JaggerJo:
> WIP [#2121](https://github.com/Live-Charts/LiveCharts2/pull/2121)

any way to support this?

beto-rodriguez:
Meanwhile you can use the NuGet provided [here](https://github.com/Live-Charts/LiveCharts2/pull/2145#issuecomment-4262920026).

V12 should be ready at NuGet soon, but you can get a working version from that feed.

beto-rodriguez:
V12 is already supported since [2.1.0-dev-292](https://www.nuget.org/packages/LiveChartsCore.SkiaSharpView.Avalonia/2.1.0-dev-292).

Celle-de:
Hi, 

[...]

PR #2143, opened 2026-04-16
Avalonia v12x

beto-rodriguez:
Updates samples to Avalonia v12, fixes breaking changes, possibly replaces #2117


beto-rodriguez:
replaced by #2145

PR #2147, opened 2026-04-17
docs: move docs references from dev to master branch

beto-rodriguez:


beto-rodriguez:
@claude apply the suggestions proposed here

PR #2154, opened 2026-04-21
Use Microsoft.WindowsAppSDK.WinUI metadata package in WinUI library

### LINKED ISSUE #2138: upgrade windowsasspdk
Please upgrade the WindowsAppSdk package to a separate WindowsAppSdk.WinUI metadata package. This will ensure a smaller binary file size when using LiveCharts.

beto-rodriguez:
Swap the full `Microsoft.WindowsAppSDK` dependency in the WinUI class library for the targeted `Microsoft.WindowsAppSDK.WinUI` metadata package. This reduces consumer binary size by not pulling in unneeded Windows App SDK components that LiveCharts does not use.

Closes #2138

### DIFF
--- a/src/skiasharp/LiveChartsCore.SkiaSharpView.WinUI/LiveChartsCore.SkiaSharpView.WinUI.csproj
+++ b/src/skiasharp/LiveChartsCore.SkiaSharpView.WinUI/LiveChartsCore.SkiaSharpView.WinUI.csproj
@@ -40,7 +40,7 @@
   </PropertyGroup>
 
   <ItemGroup>
-    <PackageReference Include="Microsoft.WindowsAppSDK" Version="1.7.250606001" />
+    <PackageReference Include="Microsoft.WindowsAppSDK.WinUI" Version="1.7.250606001" />
     <PackageReference Include="SkiaSharp.Views.WinUI" Version="$(LatestSkiaSharpVersion)" />
     <PackageReference Include="SkiaSharp.HarfBuzz" Version="$(LatestSkiaSharpVersion)" />
   </ItemGroup>


beto-rodriguez:
This PR breaks the WinUI build as-is. Do not merge.

## Root cause

The PR pins `Microsoft.WindowsAppSDK.WinUI` to version `1.7.250606001`, **but that version does not exist on NuGet**. The `.WinUI` metadata package publishes only `1.0.0-experimental1` and then jumps to `1.8.x` — there is no `1.7.x`. NuGet didn't hard-fail; it only emitted `NU1603` and silently resolved to `1.8.250906003`, which is a generation change, not a drop-in replacement.

See the NuGet index for confirmation: https://api.nuget.org/v3-flatcontainer/microsoft.windowsappsdk.winui/index.json

## Observed build failures on this branch

```
warning NU1603: LiveChartsCore.SkiaSharpView.WinUI depends on
  Microsoft.WindowsAppSDK.WinUI (>= 1.7.250606001) but
  Microsoft.WindowsAppSDK.WinUI 1.7.250606001 was not found.
  Microsoft.WindowsAppSDK.WinUI 1.8.250906003 was resolved instead.

error NETSDK1022: Duplicate 'Page' items were included.
  The duplicate items were: 'Themes\Generic.xaml'
  (both TargetFramework=net6.0-windows10.0.19041.0
   and     TargetFramework=net8.0-windows10.0.19041.0)
```

Plus a cascade of `MSB4011` "cannot be imported again" warnings because the build is simultaneously loading `Microsoft.WindowsAppSDK 1.4.230913002` (pulled in transitively, likely via `SkiaSharp.Views.WinUI`) and `Microsoft.WindowsAppSDK.WinUI 1.8.250906003`. Both metapackages import the same `Microsoft.WinUI.props/targets`, so MSBuild warns about the repeated imports.

## Why master builds and this branch doesn't

master references `Microsoft.WindowsAppSDK 1.7.250606001` (the full package, which does exist at that version). That coexists fine with the .NET SDK's default `Page` globbing, so `Themes\Generic.xaml` is included once. The 1.8 `.WinUI` metapackage adds `Page` items itself via its build targets, so the same file ends up included twice (once implicitly by the SDK, once by the package target) — hence the `NETSDK1022` duplicate.

## What would be needed for the size-reduction goal to actually land

1. Pin to a real version — e.g. `1.8.260224000` (latest 1.8 stable on NuGet at the time of writing).
2. Add `<EnableDefaultPageItems>false</EnableDefaultPageItems>` so the SDK stops auto-including `Themes\Generic.xaml`, and either rely on the package's `Page` item or declare it explicitly — but not both.
3. Investigate the `Microsoft.WindowsAppSDK 1.4.230913002` transitive (run `dotnet nuget why src/skiasharp/LiveChartsCore.SkiaSharpView.WinUI Microsoft.WindowsAppSDK`). If `SkiaSharp.Views.WinUI` still drags in the full SDK metapackage, the consumer binary-size benefit may be smaller than claimed — worth measuring pack sizes before/after to confirm the change actually delivers.

beto-rodriguez:
Recommend closing this PR. After the version bump to `1.8.250906003` (the first real published version of the `.WinUI` metapackage), the build still fails with the same `NETSDK1022` duplicate-`Page` error and `MSB4011` import conflicts. More importantly, the PR's stated goal — reducing consumer binary size — can't be achieved with this approach.

`dotnet nuget why` on this project confirms `SkiaSharp.Views.WinUI (v3.119.0)` transitively pulls in the full `Microsoft.WindowsAppSDK (v1.4.230913002)`:

```
SkiaSharp.Views.WinUI (v3.119.0)
  └── Microsoft.WindowsAppSDK (v1.4.230913002)
```

On master the direct `<PackageReference Include=\"Microsoft.WindowsAppSDK\">` overrides that transitive via NuGet's direct-wins rule, so a single generation of `Microsoft.WinUI.props/targets` gets imported and the build is clean. This PR swaps the direct dep to a different package ID (`Microsoft.WindowsAppSDK.WinUI`), so the transitive `Microsoft.WindowsAppSDK 1.4.x` is no longer overridden — both packages now sit side-by-side and both import the same props/targets → `MSB4011`.

Even if the build issues are patched (with `<EnableDefaultPageItems>false</EnableDefaultPageItems>` plus a `VersionOverride` on the transitive), consumers of `LiveChartsCore.SkiaSharpView.WinUI` would still ship the full `Microsoft.WindowsAppSDK` through the `SkiaSharp.Views.WinUI` transitive — so the size benefit this PR promises is effectively zero until `SkiaSharp.Views.WinUI` itself migrates off the full package upstream.

Suggested path forward for #2138:

1. Close this PR.
2. File an issue on mono/SkiaSharp asking `SkiaSharp.Views.WinUI` to migrate its reference from `Microsoft.WindowsAppSDK` to `Microsoft.WindowsAppSDK.WinUI`.
3. Once that lands upstream, revisit LiveCharts's own reference — at that point the swap actually becomes meaningful.

PR #2157, opened 2026-04-21
fix: trim orphan segments from path when null gaps are introduced

### LINKED ISSUE #2132: Gaps render issue
> @beto-rodriguez  thank you so much for the library. I also had this issue but since I updated to the new version the application doesn't crash. However while testing I noticed that the points are now updating (where it use the crash) but the line does not come back. In the screen recording below can see that the moment I click on Black the line stops working for the new locations (this is also the point it was crashing before version 2.0.0.0) and does not recover. Any idea's? 
> 
> https://github.com/user-attachments/assets/c6ed6386-37bd-4137-8e8d-2a898252b473
> 
> This is the code I use for updating the points array:
> ```csharp
> foreach (var trackList in trackListingsForGraph)
>                 {
>                     var point = SelectedListing.Positions.FirstOrDefault(x => x.X == trackList.X);
>                     Console.WriteLine(
>                         $"Updating listing for edition {trackList.X} and point is {point?.X?.ToString() ?? "NULL"},{point?.Y?.ToString() ?? "NULL"} setting Y to {trackList.Y?.ToString() ?? "NULL"}");
> 
>                     point?.Y = trackList.Y;
>                 }
> ``` 

 _Originally posted by @rneeft in [#1996](https://github.com/Live-Charts/LiveCharts2/issues/1996#issuecomment-4182900619)_

beto-rodriguez:
Fixes #2132.

## Root Cause

When `SplitByNullGaps` splits a series into sub-segments (e.g. `[1,2,null,4,5]` → `[[1,2],[4,5]]`), `CoreLineSeries.Invalidate` processes each sub-segment using a `VectorManager`. After the inner loop finishes for a sub-segment, `VectorManager._currentNode` pointed at the first segment of the NEXT sub-segment that hadn't been visited yet. Without cleanup, these "orphan trailing segments" remained in the first path's linked list, causing the line to incorrectly connect across the gap.

## Fix

Added `VectorManager.TrimTail()` that removes all nodes from the current position to the end of the list, called after each sub-segment is fully processed in `CoreLineSeries`, `CoreStepLineSeries`, and `CorePolarLineSeries`.

### DIFF
--- a/src/LiveChartsCore/CoreLineSeries.cs
+++ b/src/LiveChartsCore/CoreLineSeries.cs
@@ -478,6 +478,12 @@ public abstract class CoreLineSeries<TModel, TVisual, TLabel, TPathGeometry, TEr
             }
 
             if (!isSegmentEmpty) segmentI++;
+
+            // Remove any trailing segments left over from before a new gap was introduced.
+            // Without this, orphan segments from after the gap remain in the path for the
+            // sub-segment before the gap, causing the line to incorrectly cross the gap.
+            fillVector?.TrimTail();
+            strokeVector?.TrimTail();
         }
 
         var maxSegment = fillPathHelperContainer.Count > strokePathHelperContainer.Count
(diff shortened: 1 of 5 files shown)


beto-rodriguez:
Closing in favor of #2162, which rebuilds the underlying `VectorManager.AddConsecutiveSegment` logic. The `TrimTail` approach this PR took was on the right track but the AddLast branch's post-condition was off, so the first-draw case for PhBrb's XY repro ends up with the last point visible but disconnected. #2162 fixes that and includes a snapshot-test regression guard.

PR #2158, opened 2026-04-21
fix: chart not rendered behind ScrollViewer viewport (#1986)

### LINKED ISSUE #1986: The area behind the ScrollViewer viewport is not drawn
**Describe the bug**  
The content is not rendered when using `ScrollViewer` together with `TabControl`.

I don't know if this is related, but there is also no animation performed during scrolling (it was in 2.0.0-RC5.4)

**To Reproduce**
1. Create a `TabControl` with two `TabItem` elements.  
2. Add several charts inside a `ScrollViewer` in the first `TabItem`.  
3. Copy the charts to the second `TabItem`.  
4. Switch to the second tab, then return back to the first tab.  
5. Open the second tab again and scroll down — the content is not visible.  

**Expected behavior**
The content have to be visible.

**Screenshots**

Image

**Desktop (please complete the following information):**
 - OS: Windows 10 
 - Version 2.0.0-rc6.1

**Additional context**
Code to represent:
[lvccasedemo.zip](https://github.com/user-attachments/files/22388624/lvccasedemo.zip)

I would also like to know if there is a way to force rendering: 
1. To temporarily work around the problem.  
2. To trigger rendering so that it is possible to convert part of the UI containing a chart into an image when it is inside a `ScrollViewer` (`Avalonia.Skia.Helpers.DrawingContextHelperRenderAsync(SKCanvas canvas, Visual visual)`), since it is also often invisible and i need to scroll down to complete the rendering.

beto-rodriguez:
Fixes #1986

## Root cause

`IsRendering()` in `Chart.cs` compares `_lastFrameTimestamp` (set each time the canvas draws a frame) with `_lastMeasureTimeStamp` (set after measurement). When a chart is off-screen (hidden tab, scrolled out of viewport), the canvas stops drawing frames and the timestamps become equal. `IsRendering()` returns `false`, blocking all further measurements. This creates a deadlock: no measure → no draw → no new frame → timestamps stay equal → no measure. Avalonia 12’s rendering pipeline timing makes this particularly visible.

## Fix

Added a `_renderRetryScheduled` flag and a 100ms retry `Update()` when `IsRendering()` returns `false`. When the chart becomes visible again (tab switch, scroll), Avalonia renders the canvas and updates `_lastFrameTimestamp`. The pending retry then finds `IsRendering() == true` and measurement proceeds.

## Changes

- `src/LiveChartsCore/Chart.cs` — retry mechanism in `IsRendering()`
- `samples/AvaloniaSample/VisualTest/ScrollViewerVirtualization/` — new sample replicating async data + ScrollViewer scenario
- `tests/SharedUITests/CartesianChartTests.cs` — new `ScrollViewerVirtualizationRendersAsyncData` UI test

### DIFF
--- a/samples/AvaloniaSample/VisualTest/ScrollViewerVirtualization/View.axaml
+++ b/samples/AvaloniaSample/VisualTest/ScrollViewerVirtualization/View.axaml
@@ -1,0 +1,49 @@
+<UserControl
+    x:Class="AvaloniaSample.VisualTest.ScrollViewerVirtualization.View"
+    xmlns="https://github.com/avaloniaui"
+    xmlns:x="http://schemas.microsoft.com/winfx/2006/xaml"
+    xmlns:lvc="using:LiveChartsCore.SkiaSharpView.Avalonia"
+    xmlns:vm="using:AvaloniaSample.VisualTest.ScrollViewerVirtualization"
+    x:DataType="vm:ViewModel">
+
+    <!--
+        Visual test for https://github.com/Live-Charts/LiveCharts2/issues/1986
+        Charts in a TabControl + ScrollViewer where series data is assigned asynchronously.
+        Without the fix, charts remain blank even after data arrives because IsRendering()
+        blocks measurement while the canvas is not actively rendering frames.
+    -->
+
+    <TabControl x:Name="tabs">
+        <TabItem Header="page 1">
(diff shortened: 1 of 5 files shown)


beto-rodriguez:
replaced by #2170

PR #2162, opened 2026-04-23
fix: keep line path in sync when null gaps shift (#2132)

### LINKED ISSUE #2132: Gaps render issue

beto-rodriguez:
> Draft — supersedes #2157. The XY case from PhBrb's comment still degrades after many repeated inserts in the `Lines/AutoUpdateGaps` sample; opening for CI / review / further iteration before merging.

Fixes #2132.

## Root cause

`VectorManager.AddConsecutiveSegment` has three exit branches. Two of them leave `_currentNode` pointing *past* the node it just touched (Replace, Found). The third (AddLast) pointed *at* the just-appended node. A naïve \`TrimTail()\` (added in #2157) assumed the uniform post-condition and therefore **evicted the tail node it had just added** on cold draw — so a fresh chart with four `ObservablePoint`s produced a chart where the last point's geometry was visible but no line reached it.

On top of that, the original branch logic silently dropped segments whose Id fell into a gap between existing list Ids, and leaked walked-over nodes whenever it exited via AddLast or Replace-with-no-candidate. With repeated gap-shape changes (PhBrb's "click insert a few times") these orphans accumulated and the line started criss-crossing.

## Fix

Rewrite `AddConsecutiveSegment` around a single monotonic forward walk:
- remove nodes with Id ≤ `segment.Id` that are not the same instance (stale),
- inherit motion state when Ids match (`Copy()`) so animations flow,
- `AddBefore` when we overshoot,
- `AddLast` when we run off the end.

Post-condition is now uniform across all exit paths — `_currentNode` is the first node strictly after this segment, or null. That makes `TrimTail()` safe to call after each sub-segment, which lets us evict orphan trailing nodes left by a previous sub-segment layout.

Also plumb `TrimTail()` into `CoreLineSeries`, `CoreStepLineSeries`, and `CorePolarLineSeries` at the end of each sub-segment iteration.

## Test

A snapshot test (`LineSeriesTests.Issue2132_LastPointStaysConnected`) renders four `ObservablePoint`s and asserts the line connects all of them — the exact shape PhBrb reported broken on #2157. Reference PNG committed.

Full suite:
- `CoreTests` — 453 tests passing.
- `SnapshotTests` — 98 tests passing, including the new one.

## Known gap

The `Lines/AutoUpdateGaps` sample (on the `local/autoupdategaps-sample` branch locally) still shows line criss-crossing after many repeated inserts that cycle the gap structure. The fix is a clear improvement over #2157 and fixes the initial-render case, but is not yet end-to-end correct for the pathological insert-heavy scenario. Keeping as draft until that's resolved.


beto-rodriguez:
Replaced by #2164

PR #2164, opened 2026-04-23
fix: null-gap line rendering on inserts and resize (#2132)

beto-rodriguez:
## Summary

Fixes the cluster of #2132 symptoms around `LineSeries` with null-Y gaps: criss-crossing segments after inserting interpolated values, the "last point disconnected" regression, and the more recent "yellow-region swoops in from the top-left corner on resize" behavior that surfaced after the previous attempt (#2157 / the earlier #2162 iteration).

### What changed

Three small, orthogonal changes in series code and one rewrite of `VectorManager`:

1. **`VectorManager.AddConsecutiveSegment` walks from `list.First` per call, tracking `_lastTouched`.** The previous algorithm kept a persistent cursor and rewound to `list.First` on Id mismatch; the while-loop's eviction rule (`Id <= segment.Id && !refEq → stale`) then **wiped segments we had just added in the same pass** whenever the path still carried stale high-Id entries from a prior frame. The new algorithm has no cursor drift — it walks forward once per call (O(n) per insert, O(n^2) per sub-segment, trivial at real sub-segment sizes) and is structurally incapable of evicting its own just-added nodes. `TrimTail` now drops only nodes strictly after `_lastTouched`.

2. **Delete the `Commands.Count == 1` cleanup branch in all three line series.** It was meant to discard stale single-segment paths, but its implementation called `container.RemoveAt(segmentI)` and re-fetched, which **shifted every subsequent sub-segment onto the wrong path** in the container. That was the direct cause of the "yellow region collapses on resize" regression. `VectorManager.TrimTail` now drops stale remnants without touching the container, so this cleanup isn't needed.

3. **Run the pivot-init block for every new visual in `CoreLineSeries` and `CoreStepLineSeries`, not just when `isFirstDraw`.** A mid-life new visual that happens to be the first point of a brand-new sub-segment skips `Follows()` in `AddConsecutiveSegment` (because `list.Last is null`), and without the seed its motion state sits at the `Segment` default (0, 0) — so the setter below animated it from the top-left corner of the canvas. With the seed, it rises from the pivot at its own X, matching the first-draw experience. Non-first points in a sub-segment still get their motion state overridden by `Follows(previous tail)` inside `AddConsecutiveSegment`, so the "grow from previous" animation is preserved. `CorePolarLineSeries` already seeded every new visual at the chart center, so only the Count==1 removal + TrimTail wiring apply there.

### Scope

- `src/LiveChartsCore/Measure/VectorManager.cs` — rewritten.
- `src/LiveChartsCore/CoreLineSeries.cs` / `CoreStepLineSeries.cs` — Count==1 branch removed, pivot-init gate dropped, `TrimTail()` wired at end of each sub-segment.
- `src/LiveChartsCore/CorePolarLineSeries.cs` — Count==1 branch removed, `TrimTail()` wired.
- `tests/SnapshotTests/LineSeriesTests.cs` + `Snapshots/...png` — regression guard that the last point of a simple 4-point line stays connected on first draw.

Net: 4 files for the fix, 2 files for the test, 2 commits.

### Test plan

- [x] 98 snapshot tests pass locally.
- [x] 453 core tests pass locally.
- [x] Visually verified against a local `AutoUpdateGaps` repro (deterministic seeded sequence, "Insert 5x" button) across: single-click inserts, batched 5x inserts, 5x followed by 1x, and resize after each — no criss-crossing, no yellow-region collapse, no top-left swoop.
- [x] Benchmarks workflow posts a delta (expect neutral-to-slight improvement vs. current master, since the Count==1 branch's `RemoveAt` + `RemoveFromPaintTask` + `Clear` + `GetSegmentVisual` churn per stale-single-segment path is gone, and the new `AddConsecutiveSegment` does fewer list mutations in the common case).

### Not included

The `AutoUpdateGaps` Avalonia sample used to debug this is kept on a local branch and not shipped here — the snapshot test is the regression guard.


beto-rodriguez:
## Local benchmark run

Ran the full 27-benchmark suite locally on master and on the fix head, same machine, same session. CI's gate passes; local numbers tell the same story with one row right at the threshold.

### Line series (what the fix actually targets) — all neutral to faster

| Benchmark | Δ mean | Δ alloc |
|---|---:|---:|
| LineSeriesBench.Reinvalidate(1000) | +0.9% | **-7%** |
| LineSeriesBench.Reinvalidate(10000) | -1.6% | **-13%** |
| LineSeriesBench.UpdateOnePoint(1000) | -1.1% | **-7%** |
| LineSeriesBench.UpdateOnePoint(10000) | -2.7% | **-13%** |
| LineSeriesBench.FirstRender(1000) | -1.1% | -5% |
| LineSeriesBench.FirstRender(10000) | +3.4% | -6% |
| LineSeriesBench.ToggleNullGap(1000) | -5.1% | **-30%** |
| LineSeriesBench.ToggleNullGap(10000) | 🟢 **-10.1%** | **-42%** |

`ToggleNullGap` — the scenario the fix actually addresses — is ~10% faster and uses roughly half the memory at 10k points.

### Everything else — mostly within CI noise, three notable regressions

- 🟢 -13.5% on `CandlesticksSeriesBench.Reinvalidate` (unexpected win, likely noise).
- Column / Scatter / Box / Heat / Pie all within ±3%.
- 🔴 +10.0% on `StackedAreaSeriesBench.UpdateOnePoint` (exactly at the threshold).
- +7.5% on `StackedAreaSeriesBench.Reinvalidate`, +7.2% on `StepLineSeriesBench.Reinvalidate`, +5.5% / +9.0% on `PolarLineSeriesBench.Reinvalidate` / `UpdateOnePoint`. All noise-adjacent.

Memory allocations drop almost across the board (StepLine -8%, Polar -10%, StackedArea -6%). Net memory win.

### Takeaway

- On the target benchmarks (line series with gaps) the fix is neutral-to-faster and meaningfully cuts memory.
- The 5–10% slowdowns on Polar / StackedArea / StepLine are likely from the per-call `visual is null` read added to the render loop. Within the noise band on shared runners.
- Substantially better than the previous iteration of this PR, which had +15.4% on `PolarLineSeriesBench.Reinvalidate`.

PR #2209, opened 2026-05-04
chore(deps): bump Factos to 0.8.1

beto-rodriguez:
## Summary
- Bumps `FactosVersion` 0.8.0 → 0.8.1 in `Directory.Build.props` (single point of truth, propagates to all sample/test csprojs).
- Pulls in the input-event delivery fix from beto-rodriguez/Factos#16.

## Why
The Avalonia UI tests added in #2201 were passing under 0.8.0 by luck — a Factos bug in event dispatch was masking real behavior. #2208 surfaced the same bug. With 0.8.1 the dispatch is correct, so this PR re-runs CI on `master` HEAD to confirm the Avalonia suite is stable on its own merits.

If Avalonia still flakes here, the fault is in the chart-side fix from #2201 and we'll need a follow-up — not in Factos.

## Test plan
- [ ] CI green across all platforms
- [ ] Avalonia UI tests in particular pass cleanly (the suite that was flaky under 0.8.0)

### DIFF
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -34,7 +34,7 @@
     <UITesting>false</UITesting>
     <NoWarnUITests>$(NoWarn);NETSDK1198;NU1701;XC0022;XA0141;XAAADB0000;CS0067;CS0414;CS0108;CS0618;CS4014;CS8600;CS8601;CS8602;CS8603;CS8612;CS8618;CS8622;CS8625;CS8629;CS9264;</NoWarnUITests>
     <!-- Factos version https://github.com/beto-rodriguez/Factos -->
-    <FactosVersion>0.8.0</FactosVersion>
+    <FactosVersion>0.8.1</FactosVersion>
 
     <!-- Useful to toggle between using project references or nuget references -->
     <UseNuGetForSamples>false</UseNuGetForSamples>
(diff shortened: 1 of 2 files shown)


beto-rodriguez:
Update — `wpf-net10` failure was 0.8.0 lying, not the bump itself.

Diagnosis on the failed run:
- The crash `0xC0000005 at OpenTK.Platform.Windows.Wgl.DXOpenDeviceNV` happens during `ShouldLoadHardwareAcceleratedView`'s `GLWpfControl.Start` on the GHA windows-2025 runner.
- Same crash is present on **master at commit `9e460531`** (pre-#2201), with `Passed 3, Failed 0`. Factos 0.8.0 was sending `AllTestsCompleted` before the native exception unwound, masking it.
- 0.8.1 now reports the disconnect properly (`The client app disconnected before sending AllTestsCompleted; the remaining tests did not run`).
- Locally on Windows 11 with a real GPU: 13/13 pass cleanly under 0.8.1, GPU test included.

Pushed `5db5c60` to scope `TEST_HA_VIEWS` to non-CI runs (`CI=true` is auto-set by GitHub Actions; unset locally). Devs should run `wpf-net10` locally before merging WPF changes — that's the only environment where the HA path actually has a GPU to probe.

PR #2201 is **not** at fault for this — the crash predates it.

PR #2222, opened 2026-05-08
fix(wpf-templates): unblock direct Series/XAxes bindings inside DataTemplate (#1981)

### LINKED ISSUE #1981: WPF - Plot configuration is ignored when used inside a DataTemplate
**Describe the bug**
WPF: When using a Chart inside a Datatemplate Axes and Series are not used at all. This happens regardless of whether databound values or the new XAML objects are used.

**To Reproduce**
1. Create a control that uses a DataTemplate (DataGrid with DataGridTemplateColumn in my case)
2. Create a DataTemplate with a Chart inside
3. Either use DataBinding or XAML objects to assign Axes and Series
4. Chart(s) are rendered with default Axes and no Series

**Expected behavior**
Plot renders with the assigned Axes and Series

**Screenshots**
Expected chart (rendered with same data below the control)
Image
Actual chart inside the template
Image

**Desktop (please complete the following information):**
 - OS: Windwos 11
 - LiveCharts: v2.0.0-rc6

**Additional context**
Could be related to [this issue](https://github.com/beto-rodriguez/LiveCharts2/issues/1397)

beto-rodriguez:
## Summary

Fixes [#1981](https://github.com/Live-Charts/LiveCharts2/issues/1981).

WPF's \`FrameworkElementFactory\` silently drops a binding when the target \`DependencyProperty\` already holds a non-default local value. The chart constructor was setting \`Series\` / \`XAxes\` / \`YAxes\` / \`Sections\` / \`VisualElements\` to a fresh \`ObservableCollection\` in \`InitializeObservedProperties()\` via the CLR setter, so any direct binding in a \`DataTemplate\` (e.g. \`Series=\"{Binding ...}\"\`) was swallowed and the chart rendered empty.

## Approach

WPF-only — verified via Factos that Avalonia, MAUI, WinUI, and Uno do **not** reproduce #1981.

- \`UIPropertyTempaltes\` emits a lazy-init getter for collection-typed DPs (\`IEnumerable<TInterface>\`) only on the WPF template. The getter creates an \`ObservableCollection<T>\` on first access and stores it via \`SetCurrentValue\`, which preserves any attached binding.
- Shared cartesian / polar / pie \`InitializeObservedProperties\` is gated on \`#if !WPF_LVC\` so non-WPF platforms keep their existing constructor pre-init.
- \`chart.Series.Add(...)\` pre-Loaded keeps working on every platform — public API unchanged.

## Verification (Factos UI tests on Windows)

| Platform | Tests | Bug present? |
|---|---|---|
| WPF | 16/16 ✓ | yes (regression test fails when fix is reverted) |
| Avalonia | 15/15 ✓ | no |
| WinUI | 15/15 ✓ | no |
| MAUI | 14/14 ✓ | no |
| Uno | 15/15 ✓ | no |

CoreTests: 481/481 ✓.

## Test plan

- [x] WPF regression test added in \`SharedUITests/CartesianChartTests.cs\` (\`ShouldBindSeriesAndAxesInsideDataTemplate\`) and verified to fail on master without the fix
- [x] Sample (\`VisualTest/Issue1981\`) added for every XAML platform; the test runs gated on \`XAML_UI_TESTING\` so any future regression on Avalonia / MAUI / WinUI / Uno is caught in CI
- [x] CoreTests still pass on net8.0
- [x] Existing DataTemplate sample (\`VisualTest/DataTemplate\`, SeriesSource path) still renders

### DIFF
--- a/samples/ViewModelsSamples/Index.cs
+++ b/samples/ViewModelsSamples/Index.cs
@@ -122,6 +122,7 @@ public static class Index
         "VisualTest/TwoChartsOneSeries",
         "VisualTest/ReattachVisual",
         "VisualTest/DataTemplate",
+        "VisualTest/Issue1981",
         "VisualTest/Tabs",
 
         "Test/ChangeSeriesInstance",
--- a/samples/ViewModelsSamples/Index.cs
+++ b/samples/ViewModelsSamples/Index.cs
@@ -122,7 +122,6 @@ public static class Index
         "VisualTest/TwoChartsOneSeries",
         "VisualTest/ReattachVisual",
         "VisualTest/DataTemplate",
-        "VisualTest/Issue1981",
         "VisualTest/Tabs",
 
         "Test/ChangeSeriesInstance",
--- a/generators/LiveChartsGenerators/Templates/UIPropertyTempaltes.cs
+++ b/generators/LiveChartsGenerators/Templates/UIPropertyTempaltes.cs
@@ -158,6 +158,35 @@ namespace {baseTypeSymbol.ContainingNamespace};
         XamlProperty target, FrameworkTemplate template, string propertyName, string propertyType, string sanitizedPropertyType, string declaringType, string docs,
         string converter)
     {
+        // For collection-typed DPs (IEnumerable<TInterface> where TInterface != object),
+        // emit a lazy-init getter that creates an empty ObservableCollection on first
+        // access. Issue #1981: WPF's FrameworkElementFactory silently drops bindings to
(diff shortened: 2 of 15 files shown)


samples/ViewModelsSamples/Index.cs
```diff
@@ -122,6 +122,7 @@ public static class Index
         "VisualTest/TwoChartsOneSeries",
         "VisualTest/ReattachVisual",
         "VisualTest/DataTemplate",
+        "VisualTest/Issue1981",
```
beto-rodriguez: @copilot remove this change.

PR #2223, opened 2026-05-08
fix(blazor): pass container height to MotionCanvas wrapper (#1993)

### LINKED ISSUE #1993: Unable to change aspect ratio of CartesianChart in Blazor
**Describe the bug**
In Blazor WebAssembly, the aspect ratio of a CartesianChart cannot be changed.

**To Reproduce**
In a Blazor WebAssembly App, in a Razor page, add a simple CartesianChart component inside a div tag with width and height values:
`<div style="height: 150px; width: 100%;">
            <CartesianChart
                Series="series"/>
        </div>`

**Wrong behavior**
The CartesianChart is shown with a height that is too large.

Image

**Expected behavior**
The CartesianChart should be shown with a small height.

**Further Screenshots**
When analyzed in a browser, it seems that the canvas draws over the defined div section (the green area is the area where the chart should be):

Image

Bug occurs with v2.0RC6

**Workaround**
A very hacky workaround is to set the canvas height in JavaScript like:
```
window.setCanvasHeightFromDiv = (divId, height) => {
    const div = document.getElementById(divId);
    if (!div) return;

    const canvas = div.querySelector("canvas");
    if (!canvas) return;

    // Set canvas height property (important: use property, not just style)
    canvas.height = height;

    // Optional: also set CSS for layout purposes
    canvas.style.height = height + "px";
};
```
and call it in 
```
protected async override Task OnAfterRenderAsync(bool firstRender)
{
       await JS.InvokeVoidAsync("setCanvasHeightFromDiv", "yearDiv", 200);
}
```
where "yearDiv" is the id of the enclosing tag element.

beto-rodriguez:
## Summary
- A `CartesianChart` wrapped in a fixed-height container (e.g. `<div style="height: 150px">`) overflowed past the constraint instead of fitting it. Root cause: the wrapper `<div>` inside `MotionCanvas.razor` defaulted to `height: auto`, breaking the percentage-height chain to the inner `<canvas>`. Sets the wrapper to `height: 100%` so the user's constraint propagates.
- `width` is intentionally left at the block-level default — explicitly setting `width: 100%` interacted with SKCanvasView's resize observer and produced visible flicker on hover. Block-level divs already fill parent content-box width; only height was broken.
- Adds a Blazor-only repro page + UI test (`ShouldHonorFixedContainerHeight`) that asserts `ControlSize.Height ≈ 150` after first render. Verified the test fails when the fix is reverted.

Closes #1993.

## Test plan
- [x] Local browser check on `/Test/AspectRatio` — chart fits 150px wrapper, no overflow.
- [x] Local browser check on `/General/FirstChart` — unconstrained pages render unchanged.
- [x] Verified flicker-free on hover (the reason `width: 100%` was dropped).
- [x] Reverted CSS fix locally — bug returns; `ShouldHonorFixedContainerHeight` would fail.
- [x] `test-browser` job (`blazor` matrix entry) auto-runs `ShouldHonorFixedContainerHeight` in CI — verify green on this PR.

### DIFF
--- a/samples/BlazorSample/Pages/Test/AspectRatio/View.razor
+++ b/samples/BlazorSample/Pages/Test/AspectRatio/View.razor
@@ -1,0 +1,20 @@
+@page "/Test/AspectRatio"
+
+@using LiveChartsCore
+@using LiveChartsCore.SkiaSharpView
+@using LiveChartsCore.SkiaSharpView.Blazor
+
+<div style="height: 150px; width: 100%;">
+    <CartesianChart
+        @ref="Chart"
+        Series="@series">
+    </CartesianChart>
+</div>
+
+@code {
+    private ISeries[] series = [
+        new LineSeries<int> { Values = [5, 10, 8, 4] }
+    ];
+
+    public CartesianChart Chart = null!;
+}
--- a/src/skiasharp/LiveChartsCore.SkiaSharpView.Blazor/MotionCanvas.razor.css
+++ b/src/skiasharp/LiveChartsCore.SkiaSharpView.Blazor/MotionCanvas.razor.css
@@ -1,4 +1,8 @@
-﻿::deep canvas {
+﻿div {
+    height: 100%;
+}
+
+::deep canvas {
     height: 100%;
     width: 100%;
 }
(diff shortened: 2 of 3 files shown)


vincentnl78:
I feel this issue still exists.

Taking a clean BlazorApp with 2.0.5. and the canvas keeps exploding - w-full does not seem to be helping.

image

PR #2231, opened 2026-05-11
fix(uno-winui): Uno-Skia Android/iOS BindableView typeref + blank render

beto-rodriguez:
## Summary

Fixes the long-running blank-chart on Uno-Skia Android sample (`samples/UnoPlatformSample` running on `net10.0-android`). Sample now paints every chart end-to-end on a Pixel 5 API 35 emulator (Apple Silicon, Uno.Sdk 6.5.33). Same fix applies to `net10.0-ios` and `net10.0-maccatalyst`, though only Android was verified visually for this PR.

## Root cause (one mistake, two symptoms)

The chart library csproj gated `<UnoFeatures>SkiaRenderer</UnoFeatures>` behind `Condition="'$(HasOS)' != 'true'"`. `HasOS` is set by any `-android`/`-ios`/`-windows`/`-maccatalyst` TFM, so the SkiaRenderer feature was applied **only** on Uno-Native desktop/wasm builds. `-android` / `-ios` builds of the chart lib therefore compiled against the **Native Android variant** of `Uno.UI` (where `UIElement` derives from `Uno.UI.Controls.BindableView`). The source-generated DP code embeds typerefs to `BindableView` via every base-method call resolved against that compile-time hierarchy.

At runtime the sample ships Uno-Skia `Uno.UI` (no `BindableView`) — the JIT raises `TypeLoadException: Could not resolve type 'Uno.UI.Controls.BindableView'` the first time it walks any chart control's vtable. In the sample's `IndexToContentConverter`, that throw is swallowed by the binding pipeline, leaving the chart subtree silently unconstructed (the prior diagnosis of "Uno-Skia Android skips measure on UserControls" was incorrect — UserControls and Grids measure fine on Uno-Skia Android; the chart was being abandoned mid-construction).

Even with the type-load fix in isolation, `MotionCanvas`'s render-mode factory was gated `#if !HAS_OS_LVC` and so picked `CPURenderMode` on `-android`, which uses `SKXamlCanvas` — a Windows-only WinUI API absent on Uno-Skia mobile. Nothing would draw.

Both symptoms fall out from the same root: target the right `Uno.UI` variant at compile time, and route the rendering path that matches it.

## What changed

* **Commit 1** — `src/skiasharp/LiveChartsCore.SkiaSharpView.Uno.WinUI/LiveChartsCore.SkiaSharpView.Uno.WinUI.csproj`: replace `Condition="'$(HasOS)' != 'true'"` on the `SkiaRenderer` UnoFeature with `Condition="!$(TargetFramework.Contains('-windows'))"`. `-windows10.x` still compiles against the real WinUI (no Uno overlay); every other TFM lines up with the Uno-Skia runtime variant.

* **Commit 2** — adds an `LVC_UNO_SKIA` define in the same csproj for `-android`/`-ios`/`-maccatalyst` and updates `#if` guards in 8 files so the Uno-Skia render path (SkiaRenderMode2 via SKCanvasElement, Uno-Skia pointer/ticker) compiles on mobile while the Native-Android (Choreographer / MotionEvent / BindableView) and iOS-Native (CADisplayLink / UIKit gesture recognizers) partials are excluded — those continue to compile only under MAUI (LVC_UNO_SKIA is not defined on MAUI csprojs, so MAUI Android/iOS are unaffected).

* **Commits 3 + 4** — `.github/workflows/livecharts.yml`: re-enable Factos `uno / net10.0-android` in `test-android` and `uno / net10.0-ios` in `test-ios`. Both were commented in `a275efe0` because the Uno-Skia mobile sample crashed on launch; the fixes in this PR unblock them. The Android job catches future Uno-Native/Uno-Skia split regressions at PR time; the iOS job also lets us measure whether the previously-cited 3× slowdown still applies on the current Uno SDK.

## Sibling-platform builds verified clean

* Uno.WinUI chart lib `-windows10.0.19041.0`, `-desktop`, `-browserwasm` (Windows side); `-android` on Mac
* MAUI chart lib `-android`

## Test plan

- [ ] CI `test-android / uno` runs cleanly (regression guard for this exact bug)
- [ ] CI `test-ios / uno` runs cleanly (verifies Apple variant of the same fix)
- [ ] All existing Factos jobs (avalonia-*, maui, winui, wpf, winforms, blazor, eto, uno-desktop, uno-windows) stay green
- [x] Visual: `samples/UnoPlatformSample` on net10.0-android renders every Lines/Basic, Bars/Basic, and Pies/Basic sample correctly on a Pixel 5 API 35 emulator (LVC_SAMPLE=Lines/Basic shown in PR screenshot below)
- [ ] Visual: same sample on `net10.0-ios` simulator (deferred — needs a Mac runner; CI `test-ios / uno` covers this)

## Screenshot

`samples/UnoPlatformSample` on `net10.0-android`, LVC_SAMPLE=Lines/Basic, after the fix — title + two LineSeries with star/circle geometries + axes all paint:

Lines/Basic on Uno-Skia Android

(The screenshot is not committed; it lives under `.claude-android/` locally. Happy to attach via a comment if the linked path doesn't resolve.)

### DIFF
--- a/.github/workflows/livecharts.yml
+++ b/.github/workflows/livecharts.yml
@@ -388,10 +388,9 @@ jobs:
           - id: maui
             tf: net10.0-android
             workloads: maui-android
-          # ToDo: find why not running.
-          # - id: uno
-          #   tf: net10.0-android
-          #   workloads: android
+          - id: uno
+            tf: net10.0-android
+            workloads: android
     needs: [pack, prepare]
     steps:
       - name: Checkout
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -34,7 +34,7 @@
     <UITesting>false</UITesting>
     <NoWarnUITests>$(NoWarn);NETSDK1198;NU1701;XC0022;XA0141;XAAADB0000;CS0067;CS0414;CS0108;CS0618;CS4014;CS8600;CS8601;CS8602;CS8603;CS8612;CS8618;CS8622;CS8625;CS8629;CS9264;</NoWarnUITests>
     <!-- Factos version https://github.com/beto-rodriguez/Factos -->
-    <FactosVersion>0.8.1</FactosVersion>
+    <FactosVersion>0.8.3</FactosVersion>
 
     <!-- Useful to toggle between using project references or nuget references -->
     <UseNuGetForSamples>false</UseNuGetForSamples>
--- a/samples/UnoPlatformSample/UnoPlatformSample/UnoPlatformSample.csproj
+++ b/samples/UnoPlatformSample/UnoPlatformSample/UnoPlatformSample.csproj
@@ -1,4 +1,4 @@
-﻿<Project Sdk="Uno.Sdk/6.5.31">
+﻿<Project Sdk="Uno.Sdk/6.5.33">
     <PropertyGroup>
         <TargetFrameworks>net10.0-android;net10.0-ios;net10.0-windows10.0.19041.0;net10.0-browserwasm;net10.0-desktop</TargetFrameworks>
 
--- a/src/_Shared.Native/Platforms/Android/NativeTicker.cs
+++ b/src/_Shared.Native/Platforms/Android/NativeTicker.cs
@@ -20,9 +20,10 @@
 // OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
 // SOFTWARE.
 
-#if ANDROID
+#if ANDROID && !LVC_UNO_SKIA
 
 // reachable on maui android or uno android (without skia renderer)
+// LVC_UNO_SKIA excludes the Uno-Skia chart library build.
 
 using System;
 using LiveChartsCore.Motion;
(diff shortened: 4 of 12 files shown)


beto-rodriguez:
Closing this PR. Forcing `<UnoFeatures>SkiaRenderer</UnoFeatures>` on every non-Windows TFM of the chart library (the change in 71224cb5e) drops support for Uno-Native consumers on `-android` / `-ios` / `-maccatalyst` — they would no longer be able to use LiveCharts on those TFMs. That's a regression we're not willing to ship.

The original csproj convention (`Condition="'\$(HasOS)' != 'true'"` — Native variant at OS-suffixed TFMs, Skia variant at plain `netN.0`) follows Uno's documented model: <https://aka.platform.uno/skia-rendering>. For NuGet consumers, Uno's asset-resolution machinery picks the right variant based on the consumer's own `<UnoFeatures>SkiaRenderer</UnoFeatures>` setting. The original chart-blank bug was reproducible in the in-repo project-reference dev-loop because that path bypasses Uno's NuGet-time asset swap — not because the convention itself is wrong.

We'll open a fresh PR with a narrower fix that keeps the convention intact and addresses only the actual root cause. The Factos 0.8.3 fixes from beto-rodriguez/Factos#18 are independent and remain useful; the LiveCharts side can pick them up when needed.

Diagnostic artifacts kept locally for reference (Mac dev-loop steps, screenshot of the working-but-wrong fix, etc.).

PR #2233, opened 2026-05-12
fix(motion-composer): close #2029 scale-doubling on WPF tab-switch

### LINKED ISSUE #2029: In a WPF program, when the CartesianChart is contained within a TabItem, a NullReferenceException will be thrown when the TabItem is switched
**Describe the bug**
.NET 8.0 WPF Windows Application
Add a CartesianChart to a non-current TabItem of the TabControl. After the application starts, switch to the page where the CartesianChart is located and you will notice that the control is displayed abnormally (the font becomes larger and the display is blurry). When you switch back to another TabItem, the application will crash and throw the following exception:

System.NullReferenceException
  HResult=0x80004003
  Message=Object reference not set to an instance of an object.
  Source=LiveChartsCore.SkiaSharpView.WPF
  StackTrace:
   在 LiveChartsCore.SkiaSharpView.WPF.Rendering.CompositionTargetTicker.OnCompositonTargetRendering(Object sender, EventArgs e)
   在 System.Windows.Media.MediaContext.RenderMessageHandlerCore(Object resizedCompositionTarget)
   在 System.Windows.Media.MediaContext.RenderMessageHandler(Object resizedCompositionTarget)
   在 System.Windows.Threading.ExceptionWrapper.InternalRealCall(Delegate callback, Object args, Int32 numArgs)
   在 System.Windows.Threading.ExceptionWrapper.TryCatchWhen(Object source, Delegate callback, Object args, Int32 numArgs, Delegate catchHandler)
   在 System.Windows.Threading.DispatcherOperation.InvokeImpl()
   在 MS.Internal.CulturePreservingExecutionContext.CallbackWrapper(Object obj)
   在 System.Threading.ExecutionContext.RunInternal(ExecutionContext executionContext, ContextCallback callback, Object state)

```xaml
<Window x:Class="WpfLiveChartsDemo.MainWindow"
        xmlns="http://schemas.microsoft.com/winfx/2006/xaml/presentation"
        xmlns:x="http://schemas.microsoft.com/winfx/2006/xaml"
        xmlns:d="http://schemas.microsoft.com/expression/blend/2008"
        xmlns:mc="http://schemas.openxmlformats.org/markup-compatibility/2006"
        xmlns:lvc="clr-namespace:LiveChartsCore.SkiaSharpView.WPF;assembly=LiveChartsCore.SkiaSharpView.WPF"
        xmlns:local="clr-namespace:WpfLiveChartsDemo"
        mc:Ignorable="d"
        Title="MainWindow" Height="420" Width="620" Loaded="Window_Loaded">
    <Grid>
        <TabControl x:Name="tabControl" Margin="20,20,20,20">
            <TabItem Header="TabItem1">
                <Grid >
                    <lvc:CartesianChart>
                    </lvc:CartesianChart>
                </Grid>
            </TabItem>
            <TabItem Header="TabItem2">
                <Grid>
                    <lvc:CartesianChart>
                    </lvc:CartesianChart>
                </Grid>
            </TabItem>
        </TabControl>
    </Grid>
</Window>
```

**To Reproduce**

1. Click on 'TabItem2'
2. Click on 'TabItem1'
3. See error

**Screenshots**

Image
Image

**LiveCharts Version**
- LiveChartsCore.SkiaSharpView.WPF 2.0.0-rc6.1

**Desktop**
 - OS: Windows 11
 - Version: 25H2

[WpfLiveChartsDemo.zip](https://github.com/user-attachments/files/23295080/WpfLiveChartsDemo.zip)

beto-rodriguez:
## Summary
- Guards `MotionCanvasComposer.Initialize`/`Dispose` against repeated calls without an intervening Dispose/Initialize.
- Adds a CoreTests regression that fails when the guard is reverted.
- Adds a `VisualTest/Issue2029Repro` WPF sample (two empty `CartesianChart`s in a `TabControl` with auto tab-switch) covering the original report's flow end-to-end.

Closes #2029.

## Why the bug only showed up under WPF + TabControl
WPF realizes every `TabItem`'s content at load time, but the unselected tab's content sits unattached to the active `ContentPresenter`. The chart in the unselected tab fires `Loaded` at `ActualWidth=0`, and fires `Loaded` *again* — with no `Unloaded` between — once the tab is reparented into the active host.

`Initialize` re-subscribed `OnPaintSurface` on the render mode and `FrameRequest → DrawFrame` on the canvas without unsubscribing, so by the time the tab became visible every paint ran those handlers **twice** and applied the HiDPI `Canvas.Scale(dpi)` twice: a `1.75² ≈ 3` effective scale, with only the top-left quadrant visible. The reporter rightly complained that the post-#2070 build still showed the "abnormally large/blurry" first-display symptom — that was the lingering bit; this PR closes it.

Tracking a single `_isInitialized` flag inside the composer fixes it without touching any platform view's `OnLoaded`/`OnUnloaded` ordering. The flag flips back on `Dispose`, so the normal `Init → Dispose → Init` cycle other platforms use on tab switch is unchanged.

## Before / after (WPF, `VisualTest/Issue2029Repro`, screenshot at TabItem2)

| Before | After |
| --- | --- |
| Y-axis labels at ~2× size, 0–5 cropped, no X-axis visible | Full 0–10 Y and 0–10 X axes at normal label size |

## Test plan
- [x] `dotnet bin/Debug/net8.0/CoreTests.dll` — 511 / 511 pass
- [x] `dotnet bin/Debug/net10.0/SnapshotTests.dll` — 132 / 132 pass
- [x] New `MotionCanvasComposerTesting` regression: passes with fix, fails when the guard is reverted (verified)
- [x] `LVC_SAMPLE=VisualTest/Issue2029Repro LVC_SCREENSHOT=…` on WPF — chart renders correctly on the previously-hidden tab and on switch-back, no NRE

### DIFF
--- a/samples/ViewModelsSamples/Index.cs
+++ b/samples/ViewModelsSamples/Index.cs
@@ -123,7 +123,6 @@ public static class Index
         "VisualTest/ReattachVisual",
         "VisualTest/DataTemplate",
         "VisualTest/Tabs",
-        "VisualTest/Issue2029Repro",
 
         "Test/ChangeSeriesInstance",
         "Test/Dispose",
--- a/samples/ViewModelsSamples/Index.cs
+++ b/samples/ViewModelsSamples/Index.cs
@@ -123,6 +123,7 @@ public static class Index
         "VisualTest/ReattachVisual",
         "VisualTest/DataTemplate",
         "VisualTest/Tabs",
+        "VisualTest/Issue2029Repro",
 
         "Test/ChangeSeriesInstance",
         "Test/Dispose",
--- a/samples/WPFSample/VisualTest/Issue2029Repro/View.xaml
+++ b/samples/WPFSample/VisualTest/Issue2029Repro/View.xaml
@@ -1,0 +1,19 @@
+<UserControl x:Class="WPFSample.VisualTest.Issue2029Repro.View"
+             xmlns="http://schemas.microsoft.com/winfx/2006/xaml/presentation"
+             xmlns:x="http://schemas.microsoft.com/winfx/2006/xaml"
+             xmlns:lvc="clr-namespace:LiveChartsCore.SkiaSharpView.WPF;assembly=LiveChartsCore.SkiaSharpView.WPF">
+    <Grid>
+        <TabControl x:Name="tabControl" Margin="20,20,20,20">
+            <TabItem Header="TabItem1">
+                <Grid>
+                    <lvc:CartesianChart x:Name="chart1" x:FieldModifier="public"/>
+                </Grid>
+            </TabItem>
+            <TabItem Header="TabItem2">
+                <Grid>
+                    <lvc:CartesianChart x:Name="chart2" x:FieldModifier="public"/>
+                </Grid>
+            </TabItem>
+        </TabControl>
+    </Grid>
+</UserControl>
--- a/samples/WPFSample/VisualTest/Issue2029Repro/View.xaml.cs
+++ b/samples/WPFSample/VisualTest/Issue2029Repro/View.xaml.cs
@@ -1,0 +1,35 @@
+using System;
+using System.Windows.Controls;
+using System.Windows.Threading;
+
+namespace WPFSample.VisualTest.Issue2029Repro;
+
+public partial class View : UserControl
+{
+    public View()
+    {
+        InitializeComponent();
+
+        // Drive the repro automatically: TabItem1 → TabItem2 → TabItem1.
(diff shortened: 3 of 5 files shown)


samples/ViewModelsSamples/Index.cs
```diff
@@ -123,6 +123,7 @@ public static class Index
         "VisualTest/ReattachVisual",
         "VisualTest/DataTemplate",
         "VisualTest/Tabs",
+        "VisualTest/Issue2029Repro",
```
beto-rodriguez: @copilot remove this change

PR #2234, opened 2026-05-12
ci: re-enable Factos uno-android tests (bump Factos to 0.8.3)

beto-rodriguez:
## Summary

Bring the `test-android (uno, net10.0-android)` job back online in CI. It was commented out in a275efe0 with `# ToDo: find why not running` — turned out to be two bugs in Factos itself, both shipped in Factos 0.8.3 (beto-rodriguez/Factos#18). This PR picks up that version and uncomments the matrix entry.

## What the two Factos bugs were

1. **Test-runner Kestrel bound to `localhost` only.** `Constants.DEFAULT_WEBSOCKETS_URI = "http://localhost:7008"` was passed verbatim to `app.Run(...)`, so Kestrel listened on `127.0.0.1` + `::1` only (verified with `lsof -iTCP:7008` → `TCP 127.0.0.1:7008 (LISTEN)`). Emulator clients dialing the host through `10.0.2.2` couldn't reach the SignalR hub.

2. **In-app Android detection always returned `false`.** `AppController.GetIsAndroid()` was a virtual method whose base implementation called `OperatingSystem.IsAndroid()`. That API is intrinsified by the JIT/AOT to a compile-time constant determined by the *calling assembly's* `TargetFramework`. `Factos.Client.dll` targets `net10.0` (no `-android` variant), so the intrinsic folded to `false` even on a real Android device. Per-platform subclass overrides (`UnoAppController.GetIsAndroid` with `#if ANDROID return true`) didn't rescue the call site because the trimmer/JIT devirtualized the call to the base implementation. So `WebSocketsProtocolHandler.Execute` skipped the `localhost → 10.0.2.2` URI rewrite that emulator clients need.

Factos 0.8.3 binds Kestrel to all interfaces and inlines a runtime filesystem probe (`File.Exists("/system/bin/app_process")`) at the handshake call site — opaque to the trimmer, cannot be intrinsified.

## What this PR exercises

CI builds the chart libs via the existing `pack` job, then runs the sample under `UseNuGetForSamples=true` — the same NuGet asset-resolution path production consumers use. The chart lib's csproj convention (`<PropertyGroup Condition="'\$(HasOS)' != 'true'"><UnoFeatures>\$(UnoFeatures);SkiaRenderer</UnoFeatures></PropertyGroup>`) is preserved — Native variant at OS-suffixed TFMs, Skia variant at plain `netN.0` — per <https://aka.platform.uno/skia-rendering>. Uno-Native and Uno-Skia consumers both continue to work.

## Out of scope

The in-repo project-reference dev-loop on mobile Uno (i.e. running the sample directly against `<ProjectReference>` to the chart lib for `-android` / `-ios`) still renders blank. Lib DLL and sample DLL are functionally identical in both paths, but the APK's `classes.dex` + `libxamarin-app.so` differ — the divergence is in `.NET-Android`'s ACW (Android Callable Wrapper) generation between project-ref and NuGet-pack consumption. Worth a separate investigation, but doesn't block this PR because production consumers and CI both use the NuGet path.

## Test plan

- [ ] `test-android (uno, net10.0-android, android)` job runs cleanly in CI and reaches the test assertions (no infrastructure hang).
- [ ] If any test in `tests/SharedUITests/` is actually broken on Android, it surfaces now with a normal pass/fail — that's the point of having the job back.

### DIFF
--- a/tests/SharedUITests/CartesianChartTests.cs
+++ b/tests/SharedUITests/CartesianChartTests.cs
@@ -158,8 +158,16 @@ public class CartesianChartTests
     }
 #endif
 
-#if (WPF_UI_TESTING && TEST_HA_VIEWS) || MAUI_UI_TESTING || WINUI_UI_TESTING || (UNO_UI_TESTING && HAS_OS_LVC)
-    // native platforms where gpu is supported
+#if (WPF_UI_TESTING && TEST_HA_VIEWS) || MAUI_UI_TESTING || WINUI_UI_TESTING || (UNO_UI_TESTING && HAS_OS_LVC && !LVC_UNO_SKIA)
+    // native platforms where gpu is supported.
+    //
+    // The Uno arm of this gate excludes builds with the Uno SkiaRenderer
+    // feature enabled: under Uno-Skia the chart renders through
+    // Uno.WinUI.Graphics2DSK's SKCanvasElement (no native XAML
+    // hardware-accelerated element), so RendererName never contains "GPU"
+    // and this test can't satisfy the assertion. HAS_OS_LVC alone is the
+    // wrong signal for "GPU is available" now that Uno-Skia is the common
+    // setup on mobile.
 
     [AppTestMethod]
     public async Task ShouldLoadHardwareAcceleratedView()
--- a/.github/workflows/livecharts.yml
+++ b/.github/workflows/livecharts.yml
@@ -388,10 +388,9 @@ jobs:
           - id: maui
             tf: net10.0-android
             workloads: maui-android
-          # ToDo: find why not running.
-          # - id: uno
-          #   tf: net10.0-android
-          #   workloads: android
+          - id: uno
+            tf: net10.0-android
+            workloads: android
     needs: [pack, prepare]
     steps:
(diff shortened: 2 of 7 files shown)


tests/SharedUITests/CartesianChartTests.cs
```diff
@@ -158,8 +158,16 @@ public async Task ShouldHonorFixedContainerHeight()
     }
 #endif
 
-#if (WPF_UI_TESTING && TEST_HA_VIEWS) || MAUI_UI_TESTING || WINUI_UI_TESTING || (UNO_UI_TESTING && HAS_OS_LVC)
-    // native platforms where gpu is supported
+#if (WPF_UI_TESTING && TEST_HA_VIEWS) || MAUI_UI_TESTING || WINUI_UI_TESTING || (UNO_UI_TESTING && HAS_OS_LVC && !LVC_UNO_SKIA)
```
beto-rodriguez, line 161: @copilot  Extend this fix to the ShouldLoadHardwareAcceleratedView for Pie and Polar charts (not sure if PolarCharts also have an equivalent test)

PR #2238, opened 2026-05-13
fix(pie): exclude gauges from PushoutBounds aggregate (#2131)

### LINKED ISSUE #2131: PieChart - GaugeItem value at 100 will break the graph
In the PieChart control, if you set the GaugeItem value to 100, the inner graph will hover above the outergraph.

Example:
```csharp
Chart.Series = GaugeGenerator.BuildSolidGauge(
    new GaugeItem(
        100,
        series =>
        {
            series.Fill = new SolidColorPaint(InnerBorderStrokeColor.ToSkia());
            series.DataLabelsFormatter = point => $"{point.Model?.Value}%";
            series.DataLabelsSize = 18;
            series.DataLabelsPaint = new SolidColorPaint(ForegroundColor.ToSkia());
            series.DataLabelsPosition = PolarLabelsPosition.ChartCenter;
            series.InnerRadius = 38;
            series.RelativeInnerRadius = -4;
            series.MaxRadialColumnWidth = 5;
        }),
    new GaugeItem(
        GaugeItem.Background,
        series =>
        {
            series.InnerRadius = 40;
            series.Fill = new SolidColorPaint(OuterBorderStrokeColor.ToSkia());
            series.DataLabelsSize = 0;
            series.RelativeInnerRadius = 4;
            series.OuterRadiusOffset = -20;
            series.MaxRadialColumnWidth = 5;
        })
    );

// Update method invoke is needed because chart was not updating ui with new value until chart got pointer over. 
Chart.CoreChart.Update();

### PR DESCRIPTION
**Pull request by beto-rodriguez**
## Summary

Re-open of #2131. PR #2179 fixed the value-above-MaxValue overflow, but the
reporter confirmed on 2.1.0-dev-365 that the gauge still renders broken when
the chart's parent container is constrained close to the chart size (110x110
in their repro). Two-commit fix:

1. **fix/CorePieSeries-PushoutBounds**: skip `HoverPushout` / `Pushout`
   reservation for gauge series in `DataFactory.GetPieBounds`.
2. **test/snapshot**: a 110x110 solid gauge snapshot that fails without the
   fix, plus an Avalonia VisualTest sample.

Fixes #2131.

## Why

`PieSeries.HoverPushout` defaults to `20`. `DataFactory.GetPieBounds` writes
that value into the per-series `TertiaryBounds`, and `PieChartEngine.Measure`
aggregates it into a chart-wide `PushoutBounds`. `CorePieSeries.Invalidate`
then subtracts `2 * maxPushout` (40px) from the available diameter on every
series before computing radii.

In a 110x110 chart that turns `minDimension` into 70.3, so the effective
chart radius is ~35px — less than the user's requested `InnerRadius=38`. The
`MaxRadialColumnWidth` clamp at line 206 of `CorePieSeries.cs`
(`if (wc * 0.5f - stackedInnerRadius > maxRadialWidth)`) never fires,
`relativeInnerRadius` stays at `-4`, and the ring collapses to a
~1-pixel sliver at the wrong radius. The screenshot below shows the bug
exactly:

| 110x110 (bug) | 300x300 (control) |
| --- | --- |
| blue ring is a thin shrunk circle inside the gray | both rings concentric at outer rim |

Gauges occupy the full ring and don't pop out on hover (`IsHoverable=false`
for the angular path, and pushout on a full-circle ring has no visible
effect anyway), so reserving 40px of chart space for them is pure waste.
The angular-gauge setup paths in `GaugeGenerator.Build` and
`XamlGeneration.SetupAngularGaugeSeries` already zero `HoverPushout`
explicitly for the same reason — fixing it at the source in `DataFactory`
makes the convention uniform across solid and angular, XAML and code-only
gauges.

The fix is one branch in `GetPieBounds` keyed on `SeriesProperties.Gauge`,
which is set by every gauge construction path (`AsSeries`,
`SetupGaugeSeries`, `SetupBackgroundGaugeSeries`,
`SetupAngularGaugeSeries`). Non-gauge pie series and dimensional aggregation
for sibling non-gauge series are unaffected.

## Behavior change for existing gauges

Existing gauges render slightly larger now (the 40px previously reserved
becomes usable radius). Updated four existing reference snapshots that drift
by 0.10% as a consequence: `Gauge`, `GaugeMultiple`, `GaugeValueExceedsMaxValue`,
`GaugeValueExceedsRangeWithMinValue`. New regression snapshot
`GaugeInTightBounds` (110x110 with the reporter's exact configuration) fails
catastrophically without the fix.

## Test plan

- [x] Repro confirmed in Avalonia at 110x110 (blue ring collapsed to 1px sliver)
- [x] After-fix screenshot matches a scaled-down version of the 300x300 control
- [x] `dotnet build src/LiveChartsCore/LiveChartsCore.csproj` — clean
- [x] `dotnet run --project tests/SnapshotTests/SnapshotTests.csproj` — 133/133 pass
- [x] Verified `GaugeInTightBounds` fails when fix is reverted (test-catches-bug check)
- [ ] CI green

### CODE REVIEW

File: samples/AvaloniaSample/MainWindowViewModel.cs [hunk 1]
```diff
@@ -9,7 +9,8 @@ public MainWindowViewModel()
     {
         Samples = [..
             ViewModelsSamples.Index.Samples,
-            "VisualTest/Issue1986Repro"
+            "VisualTest/Issue1986Repro",
```
beto-rodriguez: @copilot  remove this and the next line.

PR #2242, opened 2026-05-13
fix(designer): skip Chart.Load when View.DesignerMode is true (#2182)

### LINKED ISSUE #2182: Visual Studio WinForms Designer fails with FileNotFoundException for SkiaSharp v2.88.0.0
**Describe the bug**
Environment

IDE: Microsoft Visual Studio
Project type: Windows Forms
Component: LiveChartsCore.SkiaSharpView.WinForms.CartesianChart

Error Summary
When opening the WinForms Designer, Visual Studio fails to create the CartesianChart component.

Error Message

Failed to create component 'CartesianChart'.

System.IO.FileNotFoundException:
Could not load file or assembly 'SkiaSharp, Version=2.88.0.0,
Culture=neutral, PublicKeyToken=0738eb9f132ed756'
or one of its dependencies.

The system cannot find the file specified.

Stack Trace

File name: 'SkiaSharp, Version=2.88.0.0, Culture=neutral,
PublicKeyToken=0738eb9f132ed756'

at LiveChartsCore.SkiaSharpView.ThemesExtensions...
at LiveChartsCore.Chart.Load()
at System.Windows.Forms.Control.AssignParent(Control value)
at System.Windows.Forms.Control.ControlCollection.Remove(Control value)
at System.Windows.Forms.Control.Dispose(Boolean disposing)
at System.Windows.Forms.ContainerControl.Dispose(Boolean disposing)
at System.ComponentModel.Component.Dispose()
at System.Drawing.Design.ToolboxItem.CreateComponentsCore(...)

**Screenshots**
Image

beto-rodriguez:
Closes #2182.

## Summary

- Adds a one-line guard at the top of `Chart.Load()` that returns when `View.DesignerMode` is true. Pre-fix the property was plumbed across every view assembly but only `GeoMapChart` consulted it — Cartesian/Pie/Polar inherited the unguarded base, so the WinForms designer host ran the full theme initializer.
- Adds a `protected virtual bool IsDesignerMode` hook on `SourceGenSKChart` so a test subclass can flip the explicit `IChartView.DesignerMode` impl, and a regression test that asserts `chart.CoreChart.IsLoaded` stays `false`. Test verified to fail on the parent commit.

## Why this stops the crash

WinForms designer host → `OnParentChanged` → `CoreChart.Load()` → `GetTheme()` → `theme.Setup(...)` runs the `OnInitialized` callbacks registered by `LiveChartsCore.SkiaSharpView.ThemesExtensions.AddDefaultTheme` → those construct `SolidColorPaint` etc. → JIT-load `SkiaSharp.dll`. Under .NET Framework strong-name binding the designer host demands exact-version `SkiaSharp, Version=2.88.0.0`, which the consumer's runtime asset graph does not satisfy → `FileNotFoundException`. Guarding `Load` short-circuits before any SkiaSharp type is touched.

## What this does NOT do

It does not produce a designer preview. SkiaSharp's `SKElement.OnRender` (WPF) and `SKControl.OnPaint` (WinForms) both `return` at the top when `DesignerProperties.GetIsInDesignMode(this)` / `DesignMode` is true (still the case on current `main` of mono/SkiaSharp). After this fix the WinForms designer stops throwing and shows an empty rectangle. Hot-reload remains the only path; Avalonia is still the only platform with a real preview because the Avalonia team ships its own design surface that loads SkiaSharp.

## Test plan

- [x] `dotnet test tests/CoreTests/CoreTests.csproj --framework net8.0` — 515/515 passed
- [x] New `ChartTests.DesignerModeShortCircuitsLoad` passes with fix, fails without (verified by `git checkout HEAD~1 -- src/LiveChartsCore/Chart.cs` then re-running)
- [x] `dotnet build src/skiasharp/LiveChartsCore.SkiaSharp.WinForms/...csproj -f net8.0-windows` clean
- [ ] Manual confirmation in VS WinForms designer (cannot automate)

### DIFF
--- a/src/LiveChartsCore/Chart.cs
+++ b/src/LiveChartsCore/Chart.cs
@@ -346,6 +346,13 @@ public abstract class Chart
     /// </summary>
     public virtual void Load()
     {
+        // At design time GetTheme below would JIT-load SkiaSharp paint types via
+        // ThemesExtensions.AddDefaultTheme initializers, which crashes the .NET
+        // Framework WinForms designer host on strong-name binding (#2182).
+        // SkiaSharp's SKElement/SKControl already short-circuit paint at design
+        // time, so there's nothing to set up.
+        if (View.DesignerMode) return;
+
         IsLoaded = true;
         _isFirstDraw = true;
         var theme = GetTheme();
--- a/src/skiasharp/LiveChartsCore.SkiaSharp/SKCharts/LiveChartsGeneratedCode/SourceGenSKChart.cs
+++ b/src/skiasharp/LiveChartsCore.SkiaSharp/SKCharts/LiveChartsGeneratedCode/SourceGenSKChart.cs
@@ -58,7 +58,14 @@ public abstract partial class SourceGenSKChart : InMemorySkiaSharpChart
         CoreChart?.Load();
     }
 
-    bool IChartView.DesignerMode => false;
+    /// <summary>
+    /// Whether this in-memory chart reports itself as running in a designer
+    /// host. Always <c>false</c> in production; test fixtures override to
+    /// exercise the design-time short-circuit in <see cref="Chart.Load"/>.
+    /// </summary>
+    protected virtual bool IsDesignerMode => false;
+
+    bool IChartView.DesignerMode => IsDesignerMode;
     bool IChartView.IsDarkMode => false;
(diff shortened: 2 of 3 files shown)


kobevn9x:
Thank you so much

PR #2281, opened 2026-05-21
fix(row-series): honor asymmetric Y error endpoints

beto-rodriguez:
## Summary

Pre-existing bug surfaced by Copilot during review of #2279 (template-method refactor). `CoreRowSeries.Invalidate` used `pe.Yi` for both endpoints of the Y error whisker, silently ignoring `pe.Yj`:

```csharp
// before
e!.YError!.X = primary + primaryScale.MeasureInPixels(pe.Yi);
e.YError.X1 = primary - primaryScale.MeasureInPixels(pe.Yi);  // <- should be pe.Yj
```

With `ErrorValue(value, yErrorI, yErrorJ)` the rendered horizontal whisker was forced symmetric — `yErrorJ` was discarded. The sibling `CoreColumnSeries` does it correctly on the Y endpoint pair.

Fix is a one-character change (`Yi` → `Yj` on the X1 endpoint).

## Commits

1. **fix**: change `pe.Yi` → `pe.Yj` on `CoreRowSeries` Y-error X1 endpoint
2. **test**: add `ErrorSeriesTests.RowWithError` snapshot baseline. Data includes `ErrorValue(70, 15, 4)` so the asymmetric whisker is visually load-bearing in the baseline — reverting the fix flips the snapshot.

Regression-test invariant verified: with the fix reverted, the new snapshot fails. With the fix in place, it passes.

## Forward port

This fix will be picked up by the in-flight refactor branch #2279 (`HorizontalBarSeries.MeasureErrorBars` carries the same bug — preserved bit-identically by design) via the standard "merge master into integration" cadence after this lands.

## Test plan
- [x] `dotnet test tests/CoreTests -f net8.0` — 636/636 passing
- [x] `dotnet test tests/SnapshotTests -f net10.0 --filter RowWithError` — passes after fix
- [x] Revert-fix smoke test — snapshot test fails as expected
- [ ] CI green


beto-rodriguez:
> Note: AI-drafted comment on Beto's behalf.

Superseded by #2282 — this branch was rooted before #2279 merged. The bug now lives in `HorizontalBarSeries.MeasureErrorBars` (formerly `CoreRowSeries.Invalidate`), and master already carries the `RowWithError` snapshot test + baseline added by the bar-template refactor. Re-rooting on current master was cleaner than reconciling the conflict.

PR #2302, opened 2026-05-26
fix(text): bare LF multiline labels render as tofu (#2266)

### LINKED ISSUE #2266: How do I implement multi-line labels?
I am having trouble figuring out how to make multi-line labels work. The chatbots are telling me to implement a custom tooltip, but it feels like overkill.

Image

The newlines aren't rendering properly right now.

Here is how they are defined in code currently.

```fsharp
    let candleSeries = CandlesticksSeries<FinancialPoint>()
    candleSeries.Values <- candleValues
    candleSeries.UpFill <- mkPaint upColor
    candleSeries.DownFill <- mkPaint downColor
    candleSeries.UpStroke <- mkStrokePaint upColor 1.0f
    candleSeries.DownStroke <- mkStrokePaint downColor 1.0f
    candleSeries.Name <- "OHLC"
    // Default tooltip formatter uses arrow/dot glyphs the bundled Skia font can't
    // render — override with plain ASCII single-line OHLC.
    candleSeries.YToolTipLabelFormatter <- fun (cp: LiveChartsCore.Kernel.ChartPoint<FinancialPoint, _, _>) ->
        let fp = cp.Model
        sprintf "O %.4f\nH %.4f\nL %.4f\nC %.4f" fp.Open fp.High fp.Low fp.Close
```

What should I do here?

beto-rodriguez:
> Drafted by Claude on Beto's behalf; he has reviewed before publishing.

## Summary

Closes #2266.

Labels containing a bare `\n` with no preceding whitespace — the pattern F# `sprintf "O %.4f\nH %.4f..."` produces — rendered every line break as a notdef tofu glyph instead of actually breaking lines. The label collapsed onto a single line.

Root cause is in `DrawingTextExtensions.Tokenize`: it appended every codepoint to its `StringBuilder` **before** checking `IsTokenBoundary`, then emitted the whole buffer (boundary char included) as one token. So `"9.4700\n"` arrived in `ShapeAndPlace` as one token; the `text == "\n"` sentinel that maps to `s_newLine` only fires for a standalone `\n`, so the glued `\n` fell through to the shaper.

Why the existing snapshot coverage missed it:

| input | tokens | result |
|---|---|---|
| `"with \r\na long"` (the existing test on Windows) | `["with ", "\r", "\n", "a ", "long"]` | works — space flushes, then `\n` lands alone |
| `"O 9.4700\nH 9.5000"` (the reporter's F# code) | `["O ", "9.4700\n", "H ", "9.5000"]` | `\n` glued → tofu, no break |

Every existing `MultilineText*` / `RenderShapedGlyphsMultiLine*` test puts a space before `Environment.NewLine`. The space flushes the buffer; on Windows the `\r` of `\r\n` then accidentally double-flushes for the `\n`. Bare-LF-no-leading-whitespace was never tested.

## Fix

In `Tokenize`:
- on `\n`: flush the buffer first, then emit `"\n"` as its own token so `ShapeAndPlace` recognises it.
- drop `\r` entirely on the way in. It only ever carried a notdef glyph and existed as an accidental flusher; with the new LF handling it's redundant and dropping it normalises CRLF cleanly.

Other whitespace keeps the existing "glue to preceding token" behavior — blast radius stays in the linebreak path.

## Commits

1. `test(snapshot): regression for bare LF multiline labels` — `TextTests.MultilineTextWithBareLineFeed` uses the reporter's exact OHLC label. Verified to fail on master alone (commit reverts cleanly to show the bug).
2. `fix(text): emit bare LF as its own token so line breaks render` — the tokenizer change plus the four pre-existing baselines refreshed for a ~0.10% subpixel shift (no more phantom `\r` blob between lines; visually identical).

## Test plan
- [x] `dotnet test tests/SnapshotTests/SnapshotTests.csproj` → 156/156 pass
- [x] `dotnet test tests/CoreTests/ --framework net8.0` → 717/717 pass
- [x] new regression test fails when fix commit is stashed, passes with it applied
- [x] visual check on the new baseline (4 clean lines, no tofu) and the refreshed RTL multiline baseline


beto-rodriguez:
@claude backport this to `release/v2.0.x` branch.

PR #2315, opened 2026-05-28
test(core): pin that bar series don't force the zero baseline into bounds

beto-rodriguez:
## Summary
- Adds two characterization tests pinning a previously-untested bounds contract: `ColumnSeries` and `RowSeries` report **data-fitted** bounds for clustered values rather than dragging the value axis to `0`.
- Follow-up to #2314 (zoom auto-fit characterization), which is now merged.

## Why
With clustered values (e.g. all `~1000`), the value axis should auto-fit tightly around the cluster so small differences stay legible. The bar still grows from `0` in pixel space, but its lower portion is simply clipped — the axis does **not** force `0` into the reported `DataBounds`. This is the contract any batched/LOD bar re-implementation must match; a naive "include 0 so the baseline shows" flattens the plot from `0..1010`. (Surfaced while building the LOD batched bar series.)

## Test plan
- [x] `ColumnSeries_ClusteredValues_DoesNotForceZeroBaseline` — Y `DataBounds` ∈ [900, 1100]
- [x] `RowSeries_ClusteredValues_DoesNotForceZeroBaseline` — X `DataBounds` ∈ [900, 1100]
- [x] `dotnet test tests/CoreTests/ --framework net8.0 --filter ZoomVisibleBoundsAutoFitTests` → 5/5 green

### DIFF
--- a/tests/CoreTests/OtherTests/ZoomVisibleBoundsAutoFitTests.cs
+++ b/tests/CoreTests/OtherTests/ZoomVisibleBoundsAutoFitTests.cs
@@ -142,4 +142,63 @@ public class ZoomVisibleBoundsAutoFitTests
         Assert.IsFalse(double.IsNaN(y.VisibleDataBounds.Min), "Y min went NaN on an empty window.");
         Assert.IsFalse(double.IsNaN(y.VisibleDataBounds.Max), "Y max went NaN on an empty window.");
     }
+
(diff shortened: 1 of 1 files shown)


beto-rodriguez:
Superseding with a more complete PR that characterizes the bounds/coordinate conventions across all cartesian series, not just column/row.

PR #2329, opened 2026-06-09
feat(axis): IAxisRenderOverride seam + GroupDates opt-in

beto-rodriguez:
> _Drafted by Claude Code on Beto's behalf; reviewed by Beto before publishing._

## What

An **axis render-override seam**, mirroring `ISeriesRenderOverride`. It lets a provider take over how a cartesian axis lays out its **separators and labels**, opted into per-axis with a new `GroupDates` flag.

The motivating use case (built on top of this, in the BackersPackage): an **adaptive, multi-level date axis** — a fine unit on the first line and a coarser context on the second (`Nov Dec Jan Feb Mar` with `2020` under `Jan`), updating as you zoom (months → days → hours…). This PR is just the OSS seam; the date renderer lives in the package.

## API

- `ICartesianAxis.GroupDates` (bool, default `false`).
- `IAxisRenderOverride.TryGroup(axis, chart, min, max, out separators, out labeler)` in `Kernel/Providers`.
- `ChartEngine.GetAxisRenderOverride(ICartesianAxis) => null` (provider hook).

When `GroupDates` is set **and** a provider supplies an override, `CoreAxis` consults it each measure for the current visible range and uses the returned separators + labeler. Otherwise nothing changes.

## Why it's loop-safe

The override returns its separators/labeler; the axis stores them in **per-measure fields cached by `(min, max)`** and reads them transiently from `GetActualLabeler` / `EnumerateSeparators`. It never writes the axis' own `CustomSeparators`/`Labeler` — doing that every frame would fire a property-change → schedule another measure → loop forever. `EnsureGrouped` runs at the start of **both** the size pass (`GetPossibleSize`) and the draw pass (`BeginMeasure`), so two-line grouped labels reserve the correct margin and draw consistently.

## Tests

- `AxisRenderOverrideTests`: the override is consulted (with the visible min/max) and its labeler used when `GroupDates` is true; never consulted when false.
- Non-breaking: **766 CoreTests + 30 axis snapshots** green (default `GroupDates=false` path unchanged).

### DIFF
--- a/src/LiveChartsCore/CoreAxis.cs
+++ b/src/LiveChartsCore/CoreAxis.cs
@@ -1176,6 +1176,8 @@ public abstract class CoreAxis<TTextGeometry, TLineGeometry>
     {
         if (!GroupDates)
         {
+            if (_groupedSeparators is not null || _groupedLabeler is not null)
+                _possibleMaxLabelsSize = null; // labels are no longer grouped → recompute the size
             _groupedSeparators = null;
             _groupedLabeler = null;
             _groupedSignature = null;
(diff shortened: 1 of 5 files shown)


beto-rodriguez:
> _Posted by Claude Code on Beto's behalf._

Addressed the review comments in `3e8c8000e`:

- **Floating-point `==`** (CodeQL): the `(min, max)` cache signature now compares with `double.Equals` — an exact cache key with no `==` operator (a miss just recomputes, no correctness impact).
- **Single-use iterator** (Copilot): the override's separators are now **materialized** when cached, so the size and draw passes re-enumerate an array, not the original iterator. Pinned by a new test (`GroupedSeparators_SingleUseIterator_AreMaterialized`).
- **Stale label size** (Copilot): `_possibleMaxLabelsSize` is reset when the grouped range changes (or grouping turns off), so two-line label height is recomputed with the new labeler.
- **Docs "every measure"** (Copilot ×2): reworded to "consulted while measuring, cached by the visible range (re-runs on zoom/pan, skipped while unchanged)" in `IAxisRenderOverride` and `ChartEngine`.
- **Unresolved cref** (Copilot): the `IAxisRenderOverride` cref in `ICartesianAxis.GroupDates` is now fully-qualified.

769 CoreTests + 30 axis snapshots green.

PR #2337, opened 2026-06-17
perf: the NativeTicker should invalidate the rendering only when the canvas is invalid

ramezgerges:
Currently, the NativeTicker will cause the UI to refresh continuously (up to the refresh rate of the screen) in WinUI and Uno Platform because CompositionTarget.Rendering's contract does that as long as you're subscribed. Instead, we should only refresh when the canvas is not IsValid.

### DIFF
--- a/src/LiveChartsCore/Motion/CoreMotionCanvas.cs
+++ b/src/LiveChartsCore/Motion/CoreMotionCanvas.cs
@@ -114,6 +114,14 @@ public class CoreMotionCanvas : IDisposable
     public event Action<CoreMotionCanvas>? Validated;
 
     /// <summary>
+    /// Occurs after a frame has been drawn (once <see cref="IsValid"/> has been updated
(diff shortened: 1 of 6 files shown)


beto-rodriguez:
Thanks for the help @ramezgerges 

> CompositionTarget.Rendering's contract does that as long as you're subscribed. Instead, we should only refresh when the canvas is not IsValid

Thanks! I thought the event was only called when the control was dirty/invalid, this will prevent unnecessary draws.

About the Uno self-paced, I think the control itself can do it, we should not use an `IFrameTicker` and instead just let the Uno control in the UI handle the frames, that is for sure an error in my initial implementation of the Uno SkiaRender, `IFrameTicker` interface is normally used to access the `Choreographer` on Android, `CADDisplayLink` on iOS, `requestAnimationFrame` on the browser, etc.
 
I created #2338, It is based on your PR, but now we don't use a ticker at all, instead the Uno control in the UI handles the frames, does that work for Uno?

PR #2381, opened 2026-07-19
Fix link to LiveCharts 2 guide in README

VikumKarunathilake:
Updated the link to the LiveCharts 2 guide.

### DIFF
--- a/README.md
+++ b/README.md
@@ -34,7 +34,7 @@ LiveCharts supports:
 - Eto Forms
 - Uwp
 
-You can also use LiveCharts 2 in a console app or on the server side by installing only the core packages. Take a look at [this guide](https://livecharts.dev/docs/wpf/2.0.4/gallery).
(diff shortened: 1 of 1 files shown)


beto-rodriguez:
> [!NOTE]
> This comment was drafted by Claude (AI) and posted by me after review.

Thanks for catching this — the old `lvcharts.com/.../2.0.0-beta.330` link was long dead.

I pushed one commit on top of yours rather than changing it: the sentence promises a console/server-side rendering guide, and the gallery page doesn't cover that. The original target does still exist on the new domain, so the link now reads:

`https://livecharts.dev/docs/wpf/2.0.4/samples.general.chartToImage#build-an-image-in-the-server-side-or-console-app`

(verified the page and the `#build-an-image-in-the-server-side-or-console-app` anchor both resolve). Your domain and version bump are untouched. Merging now — thanks again!

